PR #4, opened 2017-07-19
Fix truncate

alxvasilev:


sergiohs84:
Already merged, but without 3 commits, just one :)

PR #13, opened 2017-07-24
Avoid access to sockets from app's thread

sergiohs84:


sergiohs84:
The app calls some methods like `sendMessage()`, which directly access to websockets layer. That's not safe and could result on segmentation faults, since the socket might be invalid at certain point, like during reconnects. Better to access to libws only from the library's thread.

PR #20, opened 2017-07-26
Task/manage promises error

alber2510:
I haven't added "return err" at void function:
 - chatd::Chat::onMsgUpdated(Message* cipherMsg)
 - chatd::Chat::connect(const std::string& url)

### DIFF
--- a/src/chatClient.cpp
+++ b/src/chatClient.cpp
@@ -690,6 +690,7 @@ promise::Promise<void> Client::connect(Presence pres)
             .fail([this](const promise::Error& err)
             {
                 setConnState(kDisconnected);
+                return err;
             });
         return mConnectPromise;
     }
@@ -722,6 +723,7 @@ promise::Promise<void> Client::doConnect(Presence pres)
     .fail([this](const promise::Error& err)
     {
         setConnState(kDisconnected);
+        return err;
     });
     assert(!mHeartbeatTimer);
     mHeartbeatTimer = karere::setInterval([this]()
@@ -755,6 +757,7 @@ promise::Promise<void> Client::disconnect()
     .fail([this](const promise::Error& err)
     {
         setConnState(kDisconnected);
+        return err;
     });
     mPresencedClient.disconnect();
     return mDisconnectPromise;
@@ -1852,6 +1855,7 @@ promise::Promise<void> GroupChatRoom::decryptTitle()
         wptr.throwIfDeleted();
         KR_LOG_ERROR("Error decrypting chat title for chat %s:\n%s\nFalling back to member names.", karere::Id(chatid()).toString().c_str(), err.what());
         makeTitleFromMemberNames();
+        return err;
     });
 }
 
--- a/src/chatClient.cpp
+++ b/src/chatClient.cpp
@@ -1335,7 +1335,8 @@ void GroupChatRoom::connect()
     {
         wptr.throwIfDeleted();
         mChat->connect(mUrl);
-        decryptTitle().fail([](const promise::Error& err)
+        decryptTitle()
+        .fail([](const promise::Error& err)
         {
             KR_LOG_DEBUG("Can't decrypt chatroom title. Use a a default name");
         });
@@ -1611,7 +1612,8 @@ ChatRoom* ChatRoomList::addRoom(const mega::MegaTextChat& apiRoom)
         room = new GroupChatRoom(*this, apiRoom); //also writes it to cache
         if (client.connected())
         {
-            static_cast<GroupChatRoom*>(room)->decryptTitle().fail([](const promise::Error& err)
+            static_cast<GroupChatRoom*>(room)->decryptTitle()
+            .fail([](const promise::Error& err)
             {
                 KR_LOG_DEBUG("Can't decrypt chatroom title. Use a a default name");
             });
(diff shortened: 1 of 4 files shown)


sergiohs84:
@alber2510 maybe we should add the `return err` in a commented line for those 2 void functions, to remember it's required for the day we improve the chatd connection's tracking and/or an improved error's management for message update.

sergiohs84:
@alxvasilev please, review the cases where Alber suggest to return the error for chained promises, and feel free to add more cases you may have identified. Thanks

src/chatClient.cpp
```diff
@@ -1852,6 +1855,7 @@ promise::Promise<void> GroupChatRoom::decryptTitle()
         wptr.throwIfDeleted();
         KR_LOG_ERROR("Error decrypting chat title for chat %s:\n%s\nFalling back to member names.", karere::Id(chatid()).toString().c_str(), err.what());
         makeTitleFromMemberNames();
+        return err;
```
alxvasilev: This one is arguable, as the fail() executes a fallback (sets the title to a list of member names), but it really depends how the function is intended to be used. Having it fail and set the seems more semantically correct given the name of the function. I.e. decrypt failed -> return error. We have another small issue here - if the function returns a failed promise, it should be handled by the caller, and currently it's not, it's a fire-and-forget operation. If the function returns a failed promise, you will get an "Unhandled promise fail" warning from the promise lib.

src/chatd.cpp
```diff
@@ -423,6 +423,8 @@ Promise<void> Connection::reconnect(const std::string& url)
                     mConnectPromise.reject(err.msg(), err.code(), WS_ERRTYPE_DNS);
                     mLoginPromise.reject(err.msg(), err.code(), WS_ERRTYPE_DNS);
                 }
+
+                return err;
```
alxvasilev: No need here, the promise handler chain is not used by anything, so it doesn't matter whether the finally returned promise is. Actually, returning the error will cause an "Unhandled promise fail" warning to be logged by the promise lib.
However, I see another bug here - if the error is something else than an SDK error, it will be silently ignored. You need to change the fail() handle to something like:
```
    int errType =  (err.type() == ERRTYPE_MEGASDK) ? WS_ERRTYPE_DNS : err.type();
     mConnectPromise.reject(err.msg(), err.code(), errtype);
     mLoginPromise.reject(err.msg(), err.code(), errtype);

File: src/presenced.cpp [hunk 3]
```diff
@@ -350,6 +350,8 @@ Client::reconnect(const std::string& url)
                     mConnectPromise.reject(err.msg(), err.code(), WS_ERRTYPE_DNS);
                     mLoginPromise.reject(err.msg(), err.code(), WS_ERRTYPE_DNS);
                 }
+
+                return err;
```
alxvasilev: Same as with the chatd case, this code is the same.

src/chatClient.cpp
sergiohs84: I suggest to manage the failed promise by the caller. If the decryption of title, as you said, fails... it makes sense to return the error (despite the title will be set to the peer's names).

alber2510: I'm agree with Sergio. It's caller responsability to manage the promise-fail. GroupChatRoom::decryptTitle() doesn't know how it is going to be used

src/presenced.cpp
alber2510: I think this case is quite different from previous. This promise-fail is handled at chatClient.cpp:718

src/chatClient.cpp
sergiohs84: Then, @alber2510, you need to add the .fail() to handle the error properly in any outer function calling `decryptTitle()`, as Alex pointed above. If you have doubts, check with him.

src/presenced.cpp
sergiohs84: @alxvasilev can you check again, please?

alxvasilev: Look again at the code - the returned promise is not the chain output, it's mConnectPromise, so the chain output doesn't matter. If you insist to have the .fail() return the error, then you have to add another artificial .fail() which doesn't return an error. Otherwise you will get the "Unhandled promise fail" warning. Also, this code (as in chatd.cpp) silently ignores non-DNS errors, which is a major issue.

alxvasilev: Remember - it's normal that a .fail() doesn't return the error, but handles it. Only in cases where you need to attach further handlers for that error is it useful to return the error. Similar to a rethrowing an exception from a `catch` block.

src/chatClient.cpp
alxvasilev: Yep, you need to handle the error as @sergiohs84 suggests.

src/chatClient.cpp
```diff
@@ -2215,7 +2225,10 @@ bool GroupChatRoom::syncWithApi(const mega::MegaTextChat& chat)
         mEncryptedTitle = title;
         if (parent.client.connected())
         {
-            decryptTitle();
+            decryptTitle().fail([](const promise::Error& err)
+            {
+                KR_LOG_DEBUG("Can't decrypt chatroom title. Use a a default name");
```
alxvasilev: The convention with promises is to add the handlers on a new line each.

src/presenced.cpp
alber2510: ok, you have a better knowledge about promises. I remove it

src/chatClient.cpp
alber2510: @alxvasilev It is enough with commit abea0588d725dea1fd1d28eb9d6bb5b910d5cbc6 or is something else necesary?
I modify the code to adapt to convention

PR #28, opened 2017-07-31
receiving MSGID should not result in onMessageRejected

alxvasilev:


sergiohs84:
This patch should be discarded, the current code already manages this case. Only the apps need to refine the GUI

PR #29, opened 2017-07-31
Keep subscribed to presence for non-contacts participating on groupchats

sergiohs84:
Tests to perform:
 - User A is a contact of User B and both participate on groupchat1. User A removes B, but still sees status of B in groupchat1.
 - User B cancels his account, User A sends `DELPEER` for that user and removes the `Contact` object.

### DIFF
--- a/src/chatClient.cpp
+++ b/src/chatClient.cpp
@@ -2418,6 +2418,10 @@ void ContactList::syncWithApi(mega::MegaUserList& users)
     for (int i=0; i<size; i++)
     {
         auto& user = *users.get(i);
+        if (user.getVisibility() == ::mega::MegaUser::VISIBILITY_INACTIVE)
+        {
+            continue;
+        }
         apiUsers.insert(user.getHandle());
         addUserFromApi(user);
     }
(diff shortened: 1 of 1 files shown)


src/chatClient.cpp
```diff
@@ -2437,7 +2441,15 @@ void ContactList::syncWithApi(mega::MegaUserList& users)
 
 void ContactList::onUserAddRemove(mega::MegaUser& user)
 {
-    addUserFromApi(user);
+    if (user.getVisibility() == ::mega::MegaUser::VISIBILITY_INACTIVE)
+    {
+        auto it = this->find(user.getHandle());
+        removeUser(it);
```
alxvasilev: You need to check if iterator is valid.

src/chatClient.cpp
```diff
@@ -2437,7 +2441,18 @@ void ContactList::syncWithApi(mega::MegaUserList& users)
 
 void ContactList::onUserAddRemove(mega::MegaUser& user)
 {
-    addUserFromApi(user);
+    if (user.getVisibility() == ::mega::MegaUser::VISIBILITY_INACTIVE)
```
alxvasilev: Are you sure you don't want to add an INACTIVE user to the contactlist? I remember we wanted to have all users in the contactlist, and only show the ones that are not INACTIVE in the GUI.

sergiohs84: Inactive users should be effectively removed. We want to keep `HIDDEN` users, which mean ex-contact but still users of MEGA. An `INACTIVE` user is the one that closed his account, and therefore it's deleted from every record in the SDK (and it should do the same in karere). The `userid` becomes invalid too.

PR #35, opened 2017-08-03
WIP: Presenced login refine

alxvasilev:
Refined reconnect/login logic of presenced client. To test: connect with internet, then disable internet and wait for retries to start, then connect internet and confirm reconnect is done properly.


sergiohs84:
trigger compilation

src/presenced.cpp
```diff
@@ -281,10 +291,18 @@ Client::reconnect(const std::string& url)
...
-        return retry("presenced", [this](int no)
+        auto wptr = weakHandle();
+        if (mRetryCtrl)
         {
+            mRetryCtrl->abort();
+        }
+        mRetryCtrl.reset(createRetryController("presenced", [wptr, this](int no) -> Promise<void>
+        {
+            if (wptr.deleted())
+                return promise::Error("destroyed");
```
sergiohs84: Shouldn't we abort the retry controller in this case??

src/presenced.cpp
```diff
@@ -342,21 +356,32 @@ Client::reconnect(const std::string& url)
...
+                auto errtype = (err.type() == ERRTYPE_MEGASDK) ? WS_ERRTYPE_DNS : err.type();
+                if (!mConnectPromise.done())
+                    mConnectPromise.reject(err.msg(), err.code(), errtype);
+                if (!mLoginPromise.done())
+                    mLoginPromise.reject(err.msg(), err.code(), errtype);
             });
             
             return mConnectPromise
-            .then([this]()
+            .then([this]() -> Promise<void>
```
sergiohs84: Don't we want here the `weakptr` protection??

src/presenced.cpp
```diff
@@ -342,21 +356,32 @@ Client::reconnect(const std::string& url)
...
             {
                 mHeartbeatEnabled = true;
                 return login();
+            })
+            .fail([this](const promise::Error& err)
```
sergiohs84: And also here, the `weakptr`

src/presenced.cpp
alxvasilev: That's what we do - `mRetryCtrl->abort()`, or did I misunderstand what you mean?

src/presenced.cpp
alxvasilev: True

src/presenced.cpp
alxvasilev: Also true

src/presenced.cpp
sergiohs84: In line 302, where the `presence::Client` has been deleted, the code returns a promise. However, I've seen how heartbeats / retry controller results in new connection attempts even when the Client is already defunct.
With my comment, I'm just calling your attention to that point.

src/presenced.cpp
```diff
@@ -191,8 +193,16 @@ void Client::onSocketClose(int errcode, int errtype, const std::string& reason)
 
     if (mConnState < kLoggedIn) //tell retry controller that the connect attempt failed
     {
-        assert(!mLoginPromise.done());
-        mConnectPromise.reject(reason, errcode, errtype);
+        assert(!mConnectPromise.succeeded());
```
sergiohs84: No answers for this one? :(

alxvasilev: I needed to think a bit :) You are right, I removed the assert.

src/presenced.cpp
sergiohs84: Good news indeed! Thanks for the explanation

src/presenced.cpp
```diff
@@ -342,21 +355,36 @@ Client::reconnect(const std::string& url)
...
             {
+                if (wptr.deleted())
+                    return promise::_Void();
                 mHeartbeatEnabled = true;
                 return login();
+            })
+            .fail([wptr, this](const promise::Error& err)
+            {
+                if (wptr.deleted())
+                    return;
```
sergiohs84: Don't we need to return a `promise::Error()` here?

src/presenced.cpp
```diff
@@ -342,21 +355,36 @@ Client::reconnect(const std::string& url)
...
             })
-            .fail([this](const promise::Error& err)
+            .fail([wptr, this](const promise::Error& err)
             {
-                int errType =  (err.type() == ERRTYPE_MEGASDK) ? WS_ERRTYPE_DNS : err.type();
-                mConnectPromise.reject(err.msg(), err.code(), errType);
-                mLoginPromise.reject(err.msg(), err.code(), errType);
+                PRESENCED_LOG_ERROR("Connect error: %s", err.what());
+                if (wptr.deleted())
+                    return;
```
sergiohs84: Don't we need to return a promise::Error() here?

sergiohs84:
I got this 2 errors:
a) `WARNING: Unhandled promise fail. Error: 'Websocket error -1 on operation "connect"', type: 1, code: 1`
b) `WARNING: Unhandled promise fail. Error: 'aborted', type: 1, code: 2`

src/presenced.cpp
alxvasilev: Nop, in both places. You need to handle the error eventually, i.e. not return it further. If you don't intend to attach any further error handlers (like in this case), then you should not propagate the error further, or you will get the "unhanded promise fail" warning.

src/presenced.cpp
alxvasilev: Same as above

src/presenced.cpp
```diff
@@ -342,21 +355,36 @@ Client::reconnect(const std::string& url)
...
+                    return promise::_Void();
                 mHeartbeatEnabled = true;
                 return login();
+            })
+            .fail([wptr, this](const promise::Error& err)
+            {
+                if (wptr.deleted())
+                    return;
+                PRESENCED_LOG_ERROR("Error connecting or logging in: %s", err.what());
             });
```
alxvasilev: Weird, what's the warning? You use gcc, right?

src/presenced.cpp
```diff
@@ -342,21 +355,36 @@ Client::reconnect(const std::string& url)
...
-                mLoginPromise.reject(err.msg(), err.code(), errType);
+                PRESENCED_LOG_ERROR("Connect error: %s", err.what());
+                if (wptr.deleted())
+                    return;
+                auto errtype = (err.type() == ERRTYPE_MEGASDK) ? WS_ERRTYPE_DNS : err.type();
+                if (!mConnectPromise.done())
+                    mConnectPromise.reject(err.msg(), err.code(), errtype);
+                if (!mLoginPromise.done())
+                    mLoginPromise.reject(err.msg(), err.code(), errtype);
             });
```
alxvasilev: Weird, what's the warning? You use gcc, right?

sergiohs84: Yes, gcc. The warning says something like _non-void function should return a value_

sergiohs84:
This branch is totally outdated and won't be merged. Closing...

PR #38, opened 2017-08-08
Control best moment to send RECEIVED command

alber2510:
Previously RECEIVED command was sent for all new and old messages. Now, RECEIVED command is not sent while system is fetching and it is sent only once after received HISTDONE. It is never sent for a group-chat. For the rest of cases it is sent like before.

### DIFF
--- a/src/chatd.cpp
+++ b/src/chatd.cpp
@@ -1030,9 +1030,13 @@ void Connection::execCommand(const StaticBuffer& buf)
                 CHATD_LOG_DEBUG("%s: recv HISTDONE - history retrieval finished", ID_CSTR(chatid));
                 mClient.chats(chatid).onHistDone();
                 if (mClient.chats(chatid).getLastIdReceivedFromServer() != karere::Id()
-                        && ! mClient.chats(chatid).isGroup())
+                        && ! mClient.chats(chatid).isGroup()
+                        && (mClient.chats(chatid).getLastIdxConfirmedToServerAtHistDone() == CHATD_IDX_INVALID
+                            || mClient.chats(chatid).getLastIdxConfirmedToServerAtHistDone()
+                            < mClient.chats(chatid).getLastIdxReceivedFromServer()))
                 {
                     sendBuf(Command(OP_RECEIVED) + chatid + mClient.chats(chatid).getLastIdReceivedFromServer());
+                    mClient.chats(chatid).setLastIdxConfirmedToServerAtHistDone(mClient.chats(chatid).getLastIdxReceivedFromServer());
                 }
                 break;
             }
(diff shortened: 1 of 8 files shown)


src/chatd.cpp
```diff
@@ -2379,10 +2413,16 @@ void Chat::msgIncomingAfterDecrypt(bool isNew, bool isLocal, Message& msg, Idx i
...
-        if ((msg.userid != mClient.mUserId) &&
-           ((mLastReceivedIdx == CHATD_IDX_INVALID) || (idx > mLastReceivedIdx)))
+        if ((msg.userid != mClient.mUserId)
+                && ((mLastIdxReceivedFromServer == CHATD_IDX_INVALID) || (idx > mLastIdxReceivedFromServer))
+                && ! mClient.chats(mChatId).isGroup())
         {
-            sendCommand(Command(OP_RECEIVED) + mChatId + msgid);
+            mLastIdxReceivedFromServer = idx;
+            mLastIdReceivedFromServer = msgid;
+            if (!isFetchingFromServer())
```
sergiohs84: What if the incoming message is decrypted while it is fetching more history from server at the same time?

src/chatd.cpp
```diff
@@ -1029,6 +1029,15 @@ void Connection::execCommand(const StaticBuffer& buf)
                 READ_CHATID(0);
                 CHATD_LOG_DEBUG("%s: recv HISTDONE - history retrieval finished", ID_CSTR(chatid));
                 mClient.chats(chatid).onHistDone();
+                if (mClient.chats(chatid).getLastIdReceivedFromServer() != karere::Id()
+                        && ! mClient.chats(chatid).isGroup()
```
sergiohs84: There's a spare space between `!` and `mClient`

sergiohs84:
trigger compilation

polmr:
trigger compilation

src/chatd.cpp
```diff
@@ -2088,6 +2109,12 @@ void Chat::handleTruncate(const Message& msg, Idx idx)
                 CALL_DB(setLastReceived, 0);
             }
         }
+
+        if (mClient.isMessageReceivedConfirmationActive() && mLastIdxReceivedFromServer <= idx)
```
alxvasilev: You seem to duplicate mLastReceivedIdx and mLastReceivedId. Is there a reason for that?

alber2510: We try to manage different concepts
    - mLastReceivedIdx and mLastReceivedId are used to save the index and the handle of last message that you have sent to the peer and the peer has confirmed its reception
    - mLastIdxReceivedFromServer and mLastIdReceivedFromServer are used to save the last meesage that you have received from a peer and you have confirmed its reception

PR #39, opened 2017-08-09
Protect against repeated logouts

sergiohs84:
If karere has multiple pending SDK requests and the session becomes
invalid, a logout() attempt is done for each failed SDK request. Only
one would be necessary :)


alber2510:
This change it is not necessary because it is already protected at setInitState

PR #40, opened 2017-08-09
Check the chatd is logged in

sergiohs84:
There are certain operations that should not be performed until chatd connection is logged in (sent `JOIN`), but it wasn't enforced.
This branch also prevents resolution of promises already resolved.

### DIFF
--- a/src/chatd.cpp
+++ b/src/chatd.cpp
@@ -531,7 +531,7 @@ bool Connection::sendBuf(Buffer&& buf)
 //Copy the data to preserve the original
     auto rc = ws_send_msg_ex(mWebSocket, buf.buf(), buf.dataSize(), 1);
     buf.free(); //just in case, as it's content is xor-ed with the websock datamask so it's unusable
-    return rc;
+    return (rc == 0);
 }
 bool Chat::sendCommand(Command&& cmd)
 {
--- a/src/chatd.cpp
+++ b/src/chatd.cpp
@@ -749,8 +749,8 @@ HistSource Chat::getHistoryFromDbOrServer(unsigned count)
 
 void Chat::requestHistoryFromServer(int32_t count)
 {
-    // the connection must be established (for a JOIN + HIST), but not yet logged in
-    assert(mConnection.state() >= Connection::kStateConnected);
+    // the connection must be established, but might not be logged in yet (for a JOIN + HIST)
+    assert(mConnection.isConnected() || mConnection.isLoggedIn());
     mLastServerHistFetchCount = mLastHistDecryptCount = 0;
     mServerFetchState = (count > 0)
         ? kHistFetchingNewFromServer
--- a/src/chatd.cpp
+++ b/src/chatd.cpp
@@ -525,7 +525,7 @@ void Connection::reset() //immediate disconnect
 
 bool Connection::sendBuf(Buffer&& buf)
 {
-    if (!isLoggedIn())
+    if (!isLoggedIn() && !isConnected())
         return false;
 //WARNING: ws_send_msg_ex() is destructive to the buffer - it applies the websocket mask directly
 //Copy the data to preserve the original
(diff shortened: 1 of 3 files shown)


src/chatd.cpp
```diff
@@ -519,14 +525,13 @@ void Connection::reset() //immediate disconnect
...
-    if (!isOnline())
+    if (!isLoggedIn() && !isConnected())
         return false;
 //WARNING: ws_send_msg_ex() is destructive to the buffer - it applies the websocket mask directly
 //Copy the data to preserve the original
     auto rc = ws_send_msg_ex(mWebSocket, buf.buf(), buf.dataSize(), 1);
     buf.free(); //just in case, as it's content is xor-ed with the websock datamask so it's unusable
-    bool result = (!rc && isOnline());
-    return result;
+    return rc;
```
alxvasilev: This is not correct. `ws_send_msg_ex()` returns 0 on success and -1 or some other value on failure. The original code returns success if `rc` is 0 and the connection state didn't go to offline during the send. You should leave it like that.

src/chatd.cpp
```diff
@@ -744,7 +749,8 @@ HistSource Chat::getHistoryFromDbOrServer(unsigned count)
 
 void Chat::requestHistoryFromServer(int32_t count)
 {
-    assert(mConnection.isOnline());
+    // the connection must be established (for a JOIN + HIST), but not yet logged in
+    assert(mConnection.state() >= Connection::kStateConnected);
```
alxvasilev: To match the comment, the assert should be `mConnection.state() == Connection::kStateConnected`

src/chatd.h
```diff
@@ -302,9 +302,13 @@ class Connection: public karere::DeleteTrackable
     promise::Promise<void> mLoginPromise;
     Connection(Client& client, int shardNo): mClient(client), mShardNo(shardNo){}
     State state() { return mState; }
-    bool isOnline() const
+    bool isConnected() const
     {
-        return mState >= kStateConnected; //(mWebSocket && (ws_get_state(mWebSocket) == WS_STATE_CONNECTED));
+        return mState == kStateConnected;
```
alxvasilev: doesn't kStateOnline imply connected as well?

src/chatd.cpp
sergiohs84: Taking into account the fact that `libws` runs in one thread and the library runs in a different thread, relying on the connection state to return success or failure doesn't seem correct to me. If `libws` says the the message was sent successfully, why should we return a failure? Just because we have changed the status of the connection? 
Remember that a message already sent to the TPC/IP stack will still be sent if there's no connectivity, when it's back.
So here the point is to analyze what's better: return failure when the function succeed, but the library thinks it's offline, or rely on the result of the function blindly. What would be the consequence of each option?? @alxvasilev

alxvasilev: Actually all karere network I/O runs in the karere thread, so the send attempt is done immediately and synchronously. Only monitoring for events and timers runs in the libevent thread.
Even if you want to directly use `rc` you are not using it correctly - it's an int, which if zero means success, if nonzero means failure. You return it as a bool from a function where true means success, false - failure. So you effectively invert the result. As for ignoring the connection state, I agree, that part comes from the time when `isOnline()` checked the libws internal state instead of mState.
For safety, this function should return failure and not even try to send the packet, if we are supposed to be offline. So I would put a check at the start:
`if (!isOnline())
 {
    return;
 }
`

sergiohs84: I fixed the return value.
The protection for offline state is already in place.

src/chatd.cpp
sergiohs84: History from server is requested in 2 code paths, one when there's a connection already in place and it's logged in, but the other path is when the connection is alive but not yet logged in (as the comment states at the end of the line). So I consider the comment as correct.

PR #41, opened 2017-08-10
Management of invalid session

sergiohs84:
When the SDK is logged out or an ESID is received for any type of SDK request, automatically set the status of karere to "invalid session id", which results in a logout() from karere.


sergiohs84:
The apps should call `MegaChatApi::logout()` exclusively when the chat is disabled by the user. Otherwise, additional logout attempts may result in `EACCESS`

PR #46, opened 2017-08-22
Skip title notifications while it's being decrypted or made from member's names

alber2510:


polmr:
trigger compilation

src/chatClient.cpp
```diff
@@ -1783,7 +1788,7 @@ ChatRoomList::~ChatRoomList()
 GroupChatRoom::GroupChatRoom(ChatRoomList& parent, const mega::MegaTextChat& aChat)
 :ChatRoom(parent, aChat.getHandle(), true, aChat.getShard(),
   (chatd::Priv)aChat.getOwnPrivilege(), aChat.getCreationTime()),
-  mHasTitle(false), mRoomGui(nullptr)
+  mHasTitle(false), mRoomGui(nullptr), mTitleIsDecrypting(true)
```
sergiohs84: I think the fix should simply change `mHasTitle(false)` by `aChat.getTitle()` to properly initialize mHasTitle` when adding chatrooms from API

src/chatClient.cpp
```diff
@@ -1825,20 +1851,45 @@ ChatRoomList::~ChatRoomList()
...
+            wptr.throwIfDeleted();
+            if (!mHasTitle)
+            {
+                makeTitleFromMemberNames();
+            }
+        });
+    }
+    else
+    {
+        makeTitleFromMemberNames();
```
sergiohs84: Shouldn't we check here whether `mHasTitle` or not?
`if (!peers)` means we are alone in the groupchat, but it doesn't change the logic underneath.

src/chatClient.cpp
```diff
@@ -1508,10 +1524,15 @@ void GroupChatRoom::addMember(uint64_t userid, chatd::Priv priv, bool saveToDb)
         {
             it->second->mPriv = priv;
         }
+
+        nameMemberResolved = promise::_Void();
```
sergiohs84: Why do we use `promise::_Void()` here instead of `mPeers[userid]->nameResolved()`

src/chatClient.cpp
```diff
@@ -2293,8 +2374,13 @@ bool GroupChatRoom::syncWithApi(const mega::MegaTextChat& chat)
     }
     else
     {
-        clearTitle();
-        KR_LOG_DEBUG("Empty title received for group chat %s", Id(mChatid).toString().c_str());
+        // With this contidion some unecesary title changed notification is avoided.
```
sergiohs84: Please, check spelling of `contidion` and `unecesary`

sergiohs84: --> With this condition, some unnecessary notifications about title-changes are avoided.

src/chatClient.cpp
```diff
@@ -2293,8 +2374,13 @@ bool GroupChatRoom::syncWithApi(const mega::MegaTextChat& chat)
...
+        // We are notifing all permission changes and only group compositon change must be notified
```
sergiohs84: _We are still *notifying* title-changes for all *privilege* changes, when only group composition changes must be notified_

src/chatClient.cpp
```diff
@@ -2376,9 +2461,16 @@ GroupChatRoom::Member::Member(GroupChatRoom& aRoom, const uint64_t& user, chatd:
         if (buf && !buf->empty())
         {
             self->mEmail.assign(buf->buf(), buf->dataSize());
-            if (!self->mRoom.isInitializing() && !self->mRoom.hasTitle() && self->mName.size() <= 1)
+            if (self->mName.size() <= 1)
```
sergiohs84: Careful, here it's the handler for the attribute *email*, not related to the name. Do we want to trigger the `makeTitleFromMemberNames()` when we retrieve the emails, or only for names?

src/chatClient.cpp
alber2510: It's true, it's better retrurn always `mPeers[userid]->nameResolved()`

src/chatClient.cpp
alber2510: Yes, you are right. It's necessary add `mHasTitle`condition

src/chatClient.cpp
alber2510: We only execute this code if `self->mName.size() <= 1`. I understand that this mean that we don't have name, then the group name it is made with email

alxvasilev: Are you sure the mName.size() <= 1 check is safe while the chatroom is initializing (i.e. the code is called from the GroupChatRoom ctor)?

sergiohs84: @alxvasilev why do you think the `mName` variable could be not initialized yet? That var is part of the initialization list of the `GroupChatRoom::Member` ctor. Furthermore, the check aforementioned is in the callback for a _user attribute fetch_, so it's not called inside the ctor.
Please, check it again and let us know your reason

alxvasilev: mName is initialized to mName("\0", 1) in the initialization list, i.e. "empty". The code is in the attrib fetch callback, but it may get executed synchronously inside the constructor if the attribute is readily available in memory. That isInitializing() check is there for a reason.

sergiohs84: :thinking: I see what you mean. In the case where the name is a fail in cache and requires an API request, but the email is a hit and it's solved synchronously inside the ctor, the name would be empty because we are still fetching it.
Therefore, the best would be to serialize the attribute fetches. First, fetch the name. And later, in the callback for the name, fetch the email. That way, we ensure the name is available before checking it in the callback for the email.
@alber2510 please, do the amendments as above.
Thanks, Alex

alxvasilev:
You can serialize the attrib fetch easier if you use the promise interface to the attrib cache, like this:
```
mRoom.parent.client.userAttrCache().getAttr(user, USER_ATTR_NAME)
.then([wptr, this](Buffer* data)
{
      if (data && data->buf())
          return data;
      else
          return mRoom.parent.client.userAttrCache().getAttr(user, USER_ATTR_EMAIL);
})
.then([wptr, this](Buffer* nameOrEmail)
{
....
});
```

alxvasilev:
I did not use the promise interface in the original code, because I updated the title every time the name changes, so I needed to be notified about further updates of the attib. But now (if I udnerstand correctly) want to set the title only once and not update if on subsequent name  changes. So, you can use the more convenient promise interface.

sergiohs84:
"_want to set the title only once and not update if on subsequent name changes_"
That's not totally correct. We don't want to set the title each time a name is fetched from server at initialization, but only once the title is complete (i.e. avoid to report "name1, (...), (...)" and "name1, name2, (...)" and "name1, name2, name3", but only the latter).
However, we do want to update the title if the name of a peer changes after the title is set. I think that is not accomplished with the current code nor the old code.

alber2510:
trigger compilation

PR #47, opened 2017-08-23
Check if variable is different from NULL before useing it at marshall call

alber2510:


sergiohs84:
@alxvasilev please, take a look to this sanity checkups. Due to the marshalling, the buffer event becomes invalid between the marshall call and the execution, resulting in a crash

PR #50, opened 2017-09-07
WIP: Revamping of WebRTC with chatd signaling

sergiohs84:


src/rtcModule/webrtc.cpp
```diff
@@ -0,0 +1,2056 @@
...
+    return nullptr;
+}
+
+bool RtcModule::selectDevice(const std::string& devname,
+            const artc::DeviceList& devices, string& selected)
+{
+    if (devices.empty())
+    {
+        selected.clear();
+        return devname.empty();
```
sergiohs84: Shouldn't it return false here, regardless the device name was empty (attempt to select default device), since no device has been selected by this function.

javiserrano:
The work continues in the branch `video-calls`.

PR #51, opened 2017-09-11
Adjust chatd-connection-state monitoring in tests

sergiohs84:


sergiohs84:
trigger compilation

sergiohs84:
trigger compilation

src/chatd.cpp
```diff
@@ -431,18 +431,34 @@ Promise<void> Connection::reconnect(const std::string& url)
...
-            .fail([this](const promise::Error& err)
+            .fail([wptr, this](const promise::Error& err)
             {
+                if (wptr.deleted())
+                {
+                    CHATD_LOG_DEBUG("DNS resolution failed, but chatd client was deleted. Error: %s", err.what());
+                    return;
+                }
+
                 if (err.type() == ERRTYPE_MEGASDK)
```
alxvasilev: This will silently hang the process if (err.type() != ERRTYPE_MEGASDK), and mConnectPromise and mLoginPromise will never be resolved or rejected.

sergiohs84: Fixed!

PR #53, opened 2017-09-15
Support for inactive chats

sergiohs84:
This branch aims to:
 - Keep `Contact` objects for ex-contacts, to be in sync with data received from API.
 - Keep inactive `GroupChatRoom` objects, despite the user doesn't participate any more.
 - Keep inactive `PeerChatRoom` objects, despite the peer is not anymore a contact or even a MEGA user.
 - Allow to load history from inactive chatrooms.
 - Isolate `PeerChatRoom` from the `Contact`, since the chatroom exists regardless the existence of the user.


sergiohs84:
trigger compilation

src/chatClient.h
```diff
@@ -199,28 +199,39 @@ class PeerChatRoom: public ChatRoom
...
+    /**
+     * @brief The contact object representing the peer of the 1on1 chat.
+     * @note Returns nullptr when the 1on1 chat is with a user who canceled the account
+     */
+    Contact *contact() const { return mContact; }
 
     /** @brief The screen name of the peer */
     virtual const std::string& titleString() const;
 
+    /** @brief The screen email address of the peer */
```
alxvasilev: You can prevent duplicating the email property in the contact and 1on1 room by selectively returning here mContact->mEmail or "Deleted user"

src/chatClient.cpp
```diff
@@ -1467,11 +1458,55 @@ PeerChatRoom::~PeerChatRoom()
         chatd->leave(mChatid);
 }
 
+void PeerChatRoom::initContact(const uint64_t& peer)
+{
+    mContact = parent.client.contactList->contactFromUserId(peer);
+    mEmail = mContact ? mContact->email() : "Inactive account";
```
alxvasilev: You can prevent duplicating the email property in the contact and 1on1 room by selectively returning here mContact->mEmail or "Deleted user"

sergiohs84: I think the memory used by `mEmail` in the `ChatRoom` object is worthy, since the `Contact` may be deleted and also to speed up the access to the email.

sergiohs84:
trigger compilation

sergiohs84:
trigger compilation

sergiohs84:
Unfortunately, joining chatrooms where the own user doesn't participate anymore is not properly managed. Hence, it should be fixed before merging this branch.

polmr:
trigger compilation

polmr:
trigger compilation

PR #65, opened 2017-09-29
WIP: Implement audio-video calls at intermediate layer

alber2510:


tests/sdk_test/sdk_test.cpp
```diff
@@ -2070,19 +2069,13 @@ void MegaChatApiTest::TEST_Calls(unsigned int a1, unsigned int a2)
     ASSERT_CHAT_TEST(megaChatApi[a2]->openChatRoom(chatid, chatroomListener), "Can't open chatRoom account 1");
     loadHistory(a2, chatid, chatroomListener);
 
+    bool *audioVideoDeviceListLoaded = &requestFlagsChat[a1][MegaChatRequest::TYPE_LOAD_AUDIO_VIDEO_DEVICES]; *audioVideoDeviceListLoaded = false;
+    megaChatApi[a1]->loadAudioVideoDeviceList();
+    ASSERT_CHAT_TEST(waitForResponse(audioVideoDeviceListLoaded), "Timeout expired for receiving a call");
```
sergiohs84: Error message doesn't match the reason.

src/megachatapi.h
```diff
@@ -2630,6 +2631,14 @@ class MegaChatApi
      */
     void disableVideoCall(MegaChatHandle chatid, bool videoCall, MegaChatRequestListener *listener = NULL);
 
+    /**
+     * @brief Search all audio and video devices at the system at that moment.
+     * The associated request type with this request is MegaChatRequest::TYPE_LOAD_AUDIO_VIDEO_DEVICES
+     *
```
sergiohs84: Maybe this request could return in the request's object the new list of devices, it could be useful for the caller.

src/megachatapi.h
```diff
@@ -2630,6 +2631,14 @@ class MegaChatApi
      */
     void disableVideoCall(MegaChatHandle chatid, bool videoCall, MegaChatRequestListener *listener = NULL);
 
+    /**
+     * @brief Search all audio and video devices at the system at that moment.
```
sergiohs84: Provide a more detailed explanation, please. Like that existing list will be cleared, or that this function could be used to refresh the existing list in case you have plugged in new audio/video devices

PR #94, opened 2017-11-02
Improve the logic of network layer

sergiohs84:


src/net/websocketsIO.cpp
```diff
@@ -118,9 +118,15 @@ void WebsocketsClient::wsDisconnect(bool immediate)
     {
         return;
     }
-    
-    assert (thread_id == pthread_self());
-    ctx->wsDisconnect(immediate);
+    else if (ctx->wsIsConnected())
```
sergiohs84: These changes aim to properly control when the socket is closed by server using `libws`

src/net/websocketsIO.cpp
```diff
@@ -134,5 +140,13 @@ bool WebsocketsClient::wsIsConnected()
     }
     
     assert (thread_id == pthread_self());
-    return ctx->wsIsConnected();
+    bool ret = ctx->wsIsConnected();
```
sergiohs84: These changes (for `libwebsockets`) are focused to avoid a `wsDisconnect()` call when the socket has been closed by server. First, the disconnect attempt will attempt to gracefully disconnect by sending data over the socket, which results on an error message from the library. Second, the next `wsConnect()` call will fail in debug mode because the context (`ctx`) is still valid. So here we discard the context if no socket.

src/net/libwebsocketsIO.cpp
```diff
@@ -297,9 +297,19 @@ int LibwebsocketsClient::wsCallback(struct lws *wsi, enum lws_callback_reasons r
...
             client->disconnecting = false;
+            if (client->wsIsConnected())
+            {
+                WEBSOCKETS_LOG_DEBUG("Disconnect done by server");
+                struct lws *dwsi = client->wsi;
+                client->wsi = NULL;
+                lws_set_wsi_user(dwsi, NULL);
+            }
+            else
+            {
```
javiserrano: This code doesn't seem reachable because `wsIsConnected()` checks `client->wsi` and I think that it's only `NULL` when the `user` pointer is also `NULL` and there is an `if` above that checks if the `user` pointer is `NULL`,

I think that it would be more appropriate to use one log or the other depending on the value of `client->disconnecting` (it should be `true` before it's set to `false` above for graceful disconnects only).

Also, I would unconditionally set `client->wsi` and the `user` pointer to `NULL`.

Somethink like:
```
if (client->disconnecting)
{
    WEBSOCKETS_LOG_DEBUG("Graceful disconnect completed");
    client->disconnecting = false;
}
else
{
    WEBSOCKETS_LOG_DEBUG("Disconnect done by server");
}

struct lws *dwsi = client->wsi;
client->wsi = NULL;
lws_set_wsi_user(dwsi, NULL);
```

Of course, that is only valid of `client->wsi` can't be `NULL` here.
A safer approach could be to use the `client->disconnecting` to select the log, but still check if `client->wsi` is `NULL` before using it.

Also, checking the code of libwebsockets, it seems safe to not use `dwsi`, because it seems that `lws_set_wsi_user` can't trigger a callback, but anyway it's not bad to keep it to be protected about possible changes in that function.

So basically, I would just change the selection of the log in this snippet. The rest of proposed changes shouldn't change the behavior of this code at all.

src/net/websocketsIO.cpp
javiserrano: Checking the code of both implementations, it seems that this check doesn't have any impact on any of them.
`libws` checks if `mWebSocket` is `NULL` in both `wsIsConnected` and `wsDisconnect` before doing anything.
The same happens with `libwebsockets` and the field `wsi`.

It seems that `mWebSocket` can be different than `NULL` when the connection is closed by the server because its destruction on `LibwsClient::websockCloseCb` is placed inside a `marshallCall`. But this check wouldn't fix that problem. Instead, I think that it would be better to destroy `mWebSocket` before the `marshallCall` .

Also, I think that it's not good to have the assertion inside the `if` because it makes possible to delete `ctx` from a different thread without triggering the assert.

I see another potential problem related to this code (few lines below). The deletion of `ctx` here will do that `wsIsConnected` returns `false` during graceful disconnects. That's not ideal because that connection is still alive until the disconnection finishes, so there could be unexpected callbacks received in the app. Probably we should only delete `ctx` when we detect that the connection is finished (`ctx->wsIsConnected()` returns `false`) or after immediate disconnections, or when this object is recycled calling `wsConnect` again). Or, even better, we could have an `wsClosePrivateCb` that resets `ctx` and forwards the callback to `wsCloseCb`.

src/net/websocketsIO.cpp
javiserrano: With the changes in the callback of `libwebsockets` in this pull request, the first problem should be fixed, because the external disconnection seems to be correctly processed. 

The second problem could be fixed more reliably by deleting `ctx` there (in `wsConnect`). It would be safer because there isn't any guarantee of receiving a call to this function before a `wsConnect`. If the app manages immediate disconnects and disconnect callbacks correctly, it shouldn't be needed at all.

src/net/libwebsocketsIO.cpp
sergiohs84: I agree with you. I'll change the way we log the message to be based on the `disconnecting`, and keep the `if (client->wsIsConnected())` to set the `client->wsi = NULL`

PR #97, opened 2017-11-08
WIP: Notify for incoming call if call is video or audio call

alber2510:
Send and receive at RTCMD_CALL_REQUEST command AV flags


sergiohs84:
Outdated, solved with `CALLDATA`

PR #99, opened 2017-11-13
Send signalActivity when call is in progress

alber2510:
Avoid user becomes inactive while a call is in progress

### DIFF
--- a/src/chatClient.cpp
+++ b/src/chatClient.cpp
@@ -193,13 +193,6 @@ void Client::heartbeat()
         return;
     }
 
-#ifndef KARERE_DISABLE_WEBRTC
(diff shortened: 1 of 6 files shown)


src/chatClient.cpp
```diff
@@ -192,6 +192,14 @@ void Client::heartbeat()
         KR_LOG_WARNING("Heartbeat timer tick without being connected");
         return;
     }
+
+#ifndef KARERE_DISABLE_WEBRTC
+    if (rtc->isCallInProgress())
```
sergiohs84: I think it's better to add this checkup to `presenced::Client::autoAwayInEffect()`, in order to simply avoid the transition to the _away_ status during a call.

PR #100, opened 2017-11-14
Notify to app remote call status

alber2510:


src/megachatapi_impl.cpp
```diff
@@ -5241,6 +5278,15 @@ void MegaChatCallHandler::onLocalMediaError(const string errors)
 
 void MegaChatCallHandler::onRingOut(Id peer)
 {
+    assert(chatCall != NULL);
+    if (chatCall != NULL)
+    {
+        if (chatCall->getRemoteStatus() == rtcModule::ICall::kStateInitial)
```
sergiohs84: Is there any other possible status in this callback? if so, we would miss the notification to the app. If not, we should at least log a message (plus an assertion, maybe).

src/megachatapi_impl.cpp
```diff
@@ -5282,10 +5328,29 @@ MegaChatSessionHandler::~MegaChatSessionHandler()
...
+    assert(chatCall != NULL);
+    if (chatCall != NULL)
     {
-        MegaChatCallPrivate* chatCall = callHandler->getMegaChatCall();
-        chatCall->setRemoteAudioVideoFlags(session->receivedAv());
+        if (newState == rtcModule::ISession::kStateInProgress)
+        {
+            chatCall->setRemoteAudioVideoFlags(session->receivedAv());
+
+            if (chatCall->getRemoteStatus() == rtcModule::ICall::kStateRingIn)
```
sergiohs84: What's the purpose of this condition? in case the status is different, should we `assert` or log a warning?

src/megachatapi_impl.cpp
```diff
@@ -5282,10 +5328,29 @@ MegaChatSessionHandler::~MegaChatSessionHandler()
...
+        }
+        else if (newState == rtcModule::ISession::kStateTerminating)
+        {
+            chatCall->setRemoteStatus(rtcModule::ICall::kStateTerminating);
+        }
+        else if (newState == rtcModule::ISession::kStateDestroyed)
+        {
+            chatCall->setRemoteStatus(rtcModule::ICall::kStateDestroyed);
+        }
+
```
sergiohs84: I'd say we should simply `chatCall->setRemoteStatus(newState)` (apart from set the AV flags upon transition to `inProgress`).
If the new status is none of the expected ones, assert plus log message.

sergiohs84: According to our conversation, the session state can be other than in-progress, terminating or destroyed, so we need to trigger a `onChatCallUpdate()` only for those 3 cases that are of some interest to the apps.

src/megachatapi_impl.cpp
alber2510: The condition it is necessary to avoid notify several times Ring-In state when there are many clients

PR #101, opened 2017-11-20
Check if rtc module is initialized correctly before it is used

alber2510:
Check if start chatcall request has finished correctly

### DIFF
--- a/src/megachatapi_impl.cpp
+++ b/src/megachatapi_impl.cpp
@@ -910,7 +910,7 @@ void MegaChatApiImpl::sendPendingRequests()
         {
             if (!mClient->rtc)
             {
-                API_LOG_ERROR("WebRTC is not initialized");
+                API_LOG_ERROR("Start call - WebRTC is not initialized");
                 errorCode = MegaChatError::ERROR_ACCESS;
                 break;
             }
@@ -925,7 +925,7 @@ void MegaChatApiImpl::sendPendingRequests()
 
             if (findChatCallHandler(chatid))
             {
-                API_LOG_ERROR("One call already exists for speficied chat id");
+                API_LOG_ERROR("Start call - One call already exists for speficied chat id");
                 errorCode = MegaChatError::ERROR_EXIST;
                 break;
             }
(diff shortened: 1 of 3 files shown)


src/megachatapi_impl.cpp
```diff
@@ -2400,10 +2421,14 @@ bool MegaChatApiImpl::isMessageReceptionConfirmationActive() const
...
     std::vector<std::string> devicesVector;
-    mClient->rtc->getAudioInDevices(devicesVector);
-    sdkMutex.unlock();
+    if (mClient->rtc)
+    {
+        sdkMutex.lock();
+        mClient->rtc->getAudioInDevices(devicesVector);
+        sdkMutex.unlock();
+    }
+
```
sergiohs84: Log a message in case list is empty because rtc is not initialized

src/megachatapi_impl.cpp
```diff
@@ -2412,29 +2437,41 @@ MegaStringList *MegaChatApiImpl::getChatAudioInDevices()
...
     std::vector<std::string> devicesVector;
-    mClient->rtc->getVideoInDevices(devicesVector);
-    sdkMutex.unlock();
+    if (mClient->rtc)
+    {
+        sdkMutex.lock();
+        mClient->rtc->getVideoInDevices(devicesVector);
+        sdkMutex.unlock();
+    }
+
```
sergiohs84: Same as above

src/megachatapi_impl.cpp
```diff
@@ -2412,29 +2437,41 @@ MegaStringList *MegaChatApiImpl::getChatAudioInDevices()
...
-    sdkMutex.lock();
-    bool returnedValue = mClient->rtc->selectAudioInDevice(device);
-    sdkMutex.unlock();
+    bool returnedValue = false;
+    if (mClient->rtc)
+    {
+        sdkMutex.lock();
+        returnedValue = mClient->rtc->selectAudioInDevice(device);
+        sdkMutex.unlock();
+    }
```
sergiohs84: Same as above

src/megachatapi_impl.cpp
```diff
@@ -964,6 +971,13 @@ void MegaChatApiImpl::sendPendingRequests()
         }
         case MegaChatRequest::TYPE_HANG_CHAT_CALL:
         {
+            if (!mClient->rtc)
+            {
+                API_LOG_ERROR("WebRTC is not initialized");
+                errorCode = MegaChatError::ERROR_ACCESS;
+                break;
+            }
+
```
sergiohs84: Fix `hangupAll()` method, which destroys the call instead of calling `hangup()` for each call

PR #112, opened 2017-11-28
Release/v1.0.1

sergiohs84:
- Fix outdated unread count at startup
- Better management of disconnects
- Notify when all chatrooms are logged in
- Fix unread count when resumed from cache


sergiohs84:
trigger compilation

PR #120, opened 2017-12-09
QT bindings

javiserrano:


javiserrano:
Tested on Linux.
We should test on Windows and macOS before merging, and maybe also consider to rename base64.cpp to base64url.cpp or put its contents in another file to not cause conflicts due to the file with the same name in the SDK.

PR #124, opened 2017-12-18
Implement reject JOINRANGEHIST when un-received history is much long

alber2510:
Check manual sending

### DIFF
--- a/src/chatd.cpp
+++ b/src/chatd.cpp
@@ -849,6 +849,7 @@ Chat::Chat(Connection& conn, Id chatid, Listener* listener,
     mListener->init(*this, mDbInterface);
     CALL_CRYPTO(setUsers, &mUsers);
     assert(mDbInterface);
+    initChat();
     ChatDbInfo info;
     mDbInterface->getHistoryInfo(info);
     mOldestKnownMsgId = info.oldestDbId;
@@ -1096,6 +1097,11 @@ void Connection::execCommand(const StaticBuffer& buf)
                 else if (op == OP_JOIN)
                 {
                     chat.onJoinRejected();
+                }
+                else if (op == OP_RANGE && reason == 1)
+                {
+                    chat.clearHistory();
+                    chat.requestHistoryFromServer(-chat.initialHistoryFetchCount);
                 }
                 else
                 {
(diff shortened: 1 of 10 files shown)


src/chatd.cpp
```diff
@@ -1487,6 +1500,32 @@ void Chat::onEndCall(karere::Id userid, uint32_t clientid)
...
+    mOldestKnownMsgId = 0;
+    mLastSeenIdx = CHATD_IDX_INVALID;
+    mLastReceivedIdx = CHATD_IDX_INVALID;
+    mNextHistFetchIdx = CHATD_IDX_INVALID;
+    mLastIdReceivedFromServer = 0;
+    mLastIdxReceivedFromServer = CHATD_IDX_INVALID;
+    mLastServerHistFetchCount = 0;
+    mLastHistDecryptCount = 0;
+
+    mLastTextMsg.setState(LastTextMsgState::kNone);
```
sergiohs84: Maybe it's better to use `LastTextMsgState::clear()` here, so we also clear the `content` and `type`, not only the `state`

src/chatd.cpp
```diff
@@ -1487,6 +1500,32 @@ void Chat::onEndCall(karere::Id userid, uint32_t clientid)
...
+    mEncryptionHalted = CHATD_IDX_INVALID;
```
sergiohs84: This variable is a `boolean`, I don't think that assignment to a an invalid index is appropriate.

src/chatdDb.h
```diff
@@ -337,6 +337,12 @@ class ChatdSqliteDb: public chatd::DbInterface
         stmt.blobCol(2, buf);
         msg.assign(buf, stmt.intCol(0), stmt.uint64Col(3), stmt.intCol(1), stmt.uint64Col(4));
     }
+
+    virtual void clearHistory()
+    {
+        mDb.query("delete from history where chatid = ?", mMessages.chatId());
+        mDb.query("delete from chat_vars where chatid = ?", mMessages.chatId());
```
sergiohs84: Despite currently the `chat_vars` table only stores rows with the `name='have_all_history'`, I think it's more correct to modify this sql sentence to `delete from chat_vars where chatid = ? and name='have_all_history'`

src/megachatapi_impl.cpp
```diff
@@ -1408,6 +1408,14 @@ void MegaChatApiImpl::fireOnMessageUpdate(MegaChatMessage *msg)
     delete msg;
 }
 
+void MegaChatApiImpl::fireOnHistoryReloaded(MegaChatRoom *chat)
+{
+    for(set<MegaChatRoomListener *>::iterator it = roomListeners.begin(); it != roomListeners.end() ; it++)
+    {
+        (*it)->onHistoryReloaded(chatApi, chat);
+    }
```
sergiohs84: There is a memory leak here: `chat` should be deleted at the end of the function.

PR #132, opened 2017-12-30
Support for chatd3 (CALLDATA + REJECT)

sergiohs84:
A new chatd protocol version using a suffix for the URL is implemented here with the goal of improved videocall signaling and improved history updates.


alber2510:
It should be merged after 141

examples/qt/chatWindow.cpp
```diff
@@ -268,7 +268,6 @@ void ChatWindow::updateSeen()
 }
 void ChatWindow::onMessageEdited(const chatd::Message& msg, chatd::Idx idx)
 {
-    assert(msg.userp);
```
sergiohs84: Why do you remove this assertion?

src/rtcModule/webrtc.cpp
```diff
@@ -716,16 +816,35 @@ void Call::msgCallReqCancel(RtMessage& packet)
...
+
+    if (packet.userid != mChat.client().userId())
     {
-        return;
+        if (mState != Call::kStateReqSent && mState != Call::kStateInProgress)
+        {
+            SUB_LOG_WARNING("Ingoring unexpected CALL_REQ_DECLINE while in state %s", stateToStr(mState));
+            return;
+        }
+        if (mIsGroup || !mSessions.empty())
```
sergiohs84: Maybe a log message for this case where a peer in a group-call rejects. Improved logic to come in the future, maybe a notification about the event.
On the other hand, if there are no sessions at all...???

src/rtcModule/webrtc.cpp
```diff
@@ -716,16 +816,35 @@ void Call::msgCallReqCancel(RtMessage& packet)
...
+        {
+            return;
+        }
+        destroy(static_cast<TermCode>(TermCode::kCallRejected | TermCode::kPeer), false);
+    }
+    else // Call has been rejected by other client from same user
+    {
```
sergiohs84: Here maybe it's appropriate to have an assertion to check the clientid from received packet is different than ours.

sergiohs84:
This branch is blocked in benefit of `release/v1.1.0`, which will be merged into both `master` and `develop`. Any hotfix will be applied to the release branch.

PR #136, opened 2018-01-02
Skip reconnect event from SDK while there is a call in progress

alber2510:


sergiohs84:
Disregard this PR. A new one is created instead

PR #155, opened 2018-01-15
Release/v1.1.0

sergiohs84:
- New chatd protocol version.
- New call-signaling to enable push-notifications.
- Ability to reload local history triggered by server.
- Added Qt bindings.
- Rename file `base64` --> `base64url` (.h / .cpp) and scripts.
- Stop ringing other clients when a client rejects a call.
- Fix back references causing issues at order verification in Webclient.
- Fix the timestamp used for truncate messages.
- Update chatd's last-seen-message pointer if client is newer.
- Call arbitrage in case of collisions.


sergiohs84:
This branch include hotfixes, it must be merged into `develop` too

PR #164, opened 2018-01-22
Update develop with hotfixes

sergiohs84:


sergiohs84:
Cannot merge automatically due to conflicts. They are resolved on `task/apply-hotfixes`

PR #180, opened 2018-01-31
Adjust path for tests for execution with .pro

sergiohs84:


polmr:
trigger compilation

polmr:
trigger compilation

PR #182, opened 2018-02-02
Support for archived chats

sergiohs84:
Additionally, this branch reviews the synchronization of data received from API regarding chats: own privilege, title, peers list, peers privileges and (now) archive flag.

The corresponding implementation in Webclient is close to be ready, as they are the UI designs.

IMPORTANT: This branch modifies the DB schema just to include the `archived` flag in the `chats` table. We should force a reload of MEGAchat cache. The SDK cache should also be fully reloaded, since there might be missing inactive groupchats that require a full-reload for every user with any chat.


sergiohs84:
trigger compilation

PR #188, opened 2018-02-08
Simplify ice-server management

alber2510:


src/chatClient.cpp
```diff
@@ -834,15 +834,8 @@ promise::Promise<void> Client::doConnect(Presence pres, bool isInBackground)
...
-    .then([this, isInBackground, wptr]()
-    {
-        if (wptr.deleted())
-        {
-            return;
-        }
-        connectToChatd(isInBackground);
-    });
+    rtc->init();
+    connectToChatd(isInBackground);
```
sergiohs84: If we do it this way, we don't need the #ifdef

alber2510:
trigger compilation

PR #193, opened 2018-02-13
WIP: Take peer name in 1to1 chatroom from contact it it is possible 

alber2510:


src/megachatapi_impl.cpp
```diff
@@ -4375,17 +4375,25 @@ MegaChatRoomPrivate::MegaChatRoomPrivate(const ChatRoom &chat)
...
-        buffer = MegaChatRoomPrivate::lastnameFromBuffer(name);
-        this->peerLastnames.push_back(buffer ? buffer : "");
-        delete [] buffer;
+            buffer = MegaChatRoomPrivate::lastnameFromBuffer(name);
+            this->peerLastnames.push_back(buffer ? buffer : "");
+            delete [] buffer;
+        }
+        else
+        {
+            this->peerFirstnames.push_back(title);
```
sergiohs84: In this case, apps will miss the lastname when the 1on1 has no contact linked to it (ex-contacts and ex-users). Considering that apps may show only the firstname at certain UIs, this change will introduce a bug.
For this corner case, I suggest to whether retrieve the first/lastname directly from user-attribute's cache or to keep the title string for 1on1 with the binary format `<firstname-len><fistname><lastname>` at `PeerChatRoom` level, so the `MegaChatRoom` can extract the plain title (without the prefix) and both firstname and lastname.

PR #195, opened 2018-02-13
Fix/peer chatroom name

alber2510:


alber2510:
trigger compilation

src/chatClient.h
```diff
@@ -82,7 +82,7 @@ class ChatRoom: public chatd::Listener, public DeleteTrackable
     /** @endcond PRIVATE */
 
     /** @brief The text that will be displayed on the chat list for that chat */
-    virtual const std::string& titleString() const = 0;
+    virtual std::string titleString() const = 0;
```
sergiohs84: This change means we will be copying the `std::string` at every call without a good reason. Better to keep a reference or change to a `const char*` if the `string` object is invalid out of the scope of the function.

src/chatClient.h
```diff
@@ -223,7 +223,13 @@ class PeerChatRoom: public ChatRoom
     Contact *contact() const { return mContact; }
 
     /** @brief The screen name of the peer */
-    virtual const std::string& titleString() const;
+    virtual std::string titleString() const;
+
+    /** @brief Returns a string <fistname length><fistname><lastname>. It has binary layout
+      * First byte indicate first name length
+      */
+    virtual std::string completeTitleString() const;
```
sergiohs84: No need to create a new copy at every call. Please, keep a `const std::string &` as return datatype.

PR #207, opened 2018-02-28
Manage truncate message properly + add user's online presence at members level

jgandres:


examples/qtmegachatapi/contactItemWidget.cpp
```diff
@@ -4,23 +4,21 @@
...
+ContactItemWidget::ContactItemWidget(QWidget *parent , megachat::MegaChatApi *megaChatApi, mega::MegaApi *megaApi, mega::MegaUser *contact) :
     QWidget(parent),
     ui(new Ui::ChatItem)
 {
     mMegaApi = megaApi;
     mMegaChatApi = megaChatApi;
-    mUserHandle = userHandle;
-    const char *contactEmail = megaChatApi->getContactEmail(userHandle);
+    mUserHandle = contact->getHandle();
+    const char *contactEmail = contact->getEmail();
```
sergiohs84: Check the ownership of `MegaUser::getEmail()`. It's not yours, and you are storing a pointer to it that will eventually become invalid.

examples/qtmegachatapi/contactItemWidget.cpp
```diff
@@ -47,27 +45,29 @@ void ContactItemWidget::contextMenuEvent(QContextMenuEvent* event)
     menu.deleteLater();
 }
 
-void ContactItemWidget::updateToolTip(megachat::MegaChatHandle contactHandle)
+void ContactItemWidget::updateToolTip(mega::MegaUser *contact)
 {
    QString text = NULL;
-   char *email = mMegaChatApi->getContactEmail(contactHandle);
-   mega::MegaUser *contact = mMegaApi->getContact(email);
+   const char *email = contact->getEmail();
```
sergiohs84: Same than above, the ownership is kept by the `MegaUser`

examples/qtmegachatapi/chatWindow.cpp
```diff
@@ -145,9 +145,22 @@ void ChatWindow::onMessageUpdate(megachat::MegaChatApi* api, megachat::MegaChatM
...
-            chatMessage->setMessageContent(msg->getContent());
-            if (msg->isEdited())
-                chatMessage->markAsEdited();
+            if (msg->getType() == megachat::MegaChatMessage::TYPE_TRUNCATE)
+            {
+                truncateChatUI();
+                megachat::MegaChatMessage *auxMsg = msg->copy();
+                addMsgWidget(auxMsg, loadedMessages);
+                mChatItemWidget->setOlderMessageLoaded(msg->getMsgId());
+                mMegaChatApi->setMessageSeen(mChatRoom->getChatId(), msg->getMsgId());
```
sergiohs84: In case the truncate is ours, the app should NOT set the message as seen. It doesn't have any consequence, since MEGAchat already checks the `userid`, but better to check at the app.

examples/qtmegachatapi/MegaChatApplication.cpp
```diff
@@ -185,7 +185,10 @@ void MegaChatApplication::onUsersUpdate(mega::MegaApi * api, mega::MegaUserList
             itContacts = this->mMainWin->contactWidgets.find(userHandle);
             if (itContacts == this->mMainWin->contactWidgets.end())
             {
-                mMainWin->addContact(userHandle);
+                if (user)
```
sergiohs84: You shouldn't need this check, you're iterating a list, items of that list should be always valid.

PR #210, opened 2018-03-01
Roaming with new call

alber2510:


src/rtcModule/webrtc.cpp
```diff
@@ -2872,6 +3065,14 @@ Promise<void> Session::terminateAndDestroy(TermCode code, const std::string& msg
             return pms;
         }
     }
+
+    unsigned timeout = 1000;
```
sergiohs84: Define a constant for `1000`, something like `kSessionRetryTimeout`

src/rtcModule/webrtc.cpp
```diff
@@ -585,6 +607,53 @@ void RtcModule::hangupAll(TermCode code)
     }
 }
 
+void RtcModule::launchCallRetry(Id chatid, AvFlags av, bool starter)
```
sergiohs84: It would be a good idea for readability to refactor `starter` --> `isActiveRetry`

src/rtcModule/webrtc.cpp
```diff
@@ -775,30 +844,35 @@ void RtcModule::updatePeerAvState(Id chatid, Id callid, Id userid, uint32_t clie
...
         if (itHandler != mCallHandlers.end())
         {
-            Chat &chat = mKarereClient.mChatdClient->chats(chatid);
-            if (keepCallHandler && itHandler->second->callParticipants() && chat.isGroup())
+            if (retry || itHandler->second->callParticipants())
             {
                 itHandler->second->removeAllParticipants();
             }
-            else
+            else if (mRetryCall.find(chatid) == mRetryCall.end())
```
sergiohs84: Better to add the new condition as an `assert()` in the `else` block.

alber2510: It's necessary keep the if condition to avoid destroy the call handler during reconnection

src/rtcModule/webrtc.cpp
```diff
@@ -2745,8 +2930,16 @@ Promise<void> Session::sendOffer()
     })
     .then([wptr, this]()
     {
-        if (wptr.deleted())
+        if (wptr.deleted() || mState != Session::kStateWaitSdpAnswer)
```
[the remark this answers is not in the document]
alber2510: I've replaced the if by an assert

src/rtcModule/webrtc.cpp
sergiohs84: In that case, since it doesn't seem obvious, maybe a comment would be helpful.

PR #215, opened 2018-03-04
 Add function to enable chat

alber2510:
Replace init(no chache) + fetchnodes

### DIFF
--- a/src/chatClient.cpp
+++ b/src/chatClient.cpp
@@ -594,23 +594,25 @@ void Client::onRequestFinish(::mega::MegaApi* apiObj, ::mega::MegaRequest *reque
     case ::mega::MegaRequest::TYPE_FETCH_NODES:
     {
         api.sdk.pauseActionPackets();
-        auto state = mInitState;
-        char* pscsn = api.sdk.getSequenceNumber();
-        std::string scsn;
-        if (pscsn)
-        {
-            scsn = pscsn;
-            delete[] pscsn;
-        }
-        std::shared_ptr<::mega::MegaUserList> contactList(api.sdk.getContacts());
-        std::shared_ptr<::mega::MegaTextChatList> chatList(api.sdk.getChatList());
 
         auto wptr = weakHandle();
-        marshallCall([wptr, this, state, scsn, contactList, chatList]()
+        marshallCall([wptr, this]()
         {
             if (wptr.deleted())
                 return;
 
+            char* pscsn = api.sdk.getSequenceNumber();
+            std::string scsn;
+            if (pscsn)
+            {
+                scsn = pscsn;
+                delete[] pscsn;
+            }
+
+            std::shared_ptr<::mega::MegaUserList> contactList(api.sdk.getContacts());
+            std::shared_ptr<::mega::MegaTextChatList> chatList(api.sdk.getChatList());
+
+            auto state = mInitState;
             if (state == kInitHasOfflineSession)
             {
 // disable this safety checkup, since dumpSession() differs from first-time login value
@@ -629,20 +631,9 @@ void Client::onRequestFinish(::mega::MegaApi* apiObj, ::mega::MegaRequest *reque
             }
             else if (state == kInitWaitingNewSession || state == kInitErrNoCache)
             {
-                std::unique_ptr<char[]> sid(api.sdk.dumpSession());
-                assert(sid);
-                initWithNewSession(sid.get(), scsn, contactList, chatList)
-                .fail([this](const promise::Error& err)
-                {
-                    mSessionReadyPromise.reject(err);
-                    return err;
-                })
-                .then([this]()
-                {
-                    setInitState(kInitHasOnlineSession);
-                    mSessionReadyPromise.resolve();
-                });
+                initCacheFromSdkSession(scsn, contactList, chatList);
             }
+
             api.sdk.resumeActionPackets();
         }, appCtx);
         break;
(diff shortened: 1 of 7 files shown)


src/megachatapi.h
```diff
@@ -1562,6 +1557,16 @@ class MegaChatApi
      */
     int getInitState();
 
+    /**
+     * @brief This function enable the chat for the provided session. Replace behavior of MegaChatApi::init()
+     * (returned value MegaChatApi::INIT_NO_CACHE) + fetchnodes in order to re-create a new cache from scratch.
+     * Megachat's chache will be regenerated based on data from SDK's cache. No need to invalidate the SDK's cache.
```
sergiohs84: Megachat --> MEGAchat
chache --> cache

src/megachatapi_impl.cpp
```diff
@@ -1175,6 +1175,44 @@ int MegaChatApiImpl::getInitState()
...
 }
 
+bool MegaChatApiImpl::enableChat(const char *sid)
+{
+    if (sid == NULL)
+    {
+        API_LOG_ERROR("Failure to enable chat, invalid session");
+        return false;
+    }
+
```
sergiohs84: If init state is already a valid one, we can safely return true (chat is enabled and working)

src/megachatapi_impl.cpp
```diff
@@ -1175,6 +1175,44 @@ int MegaChatApiImpl::getInitState()
     return initState;
 }
 
+bool MegaChatApiImpl::enableChat(const char *sid)
+{
+    if (sid == NULL)
```
sergiohs84: Consider to check `sid[0] != '/0'`. Maybe it's a good protection, I've seen an `assert(sid.empty())` in `openDb()`

src/chatClient.cpp
```diff
@@ -2941,6 +2932,23 @@ const char* Client::connStateToStr(ConnState state)
     }
 }
 
+void Client::initCacheFromSdkSession(const std::string scsn, std::shared_ptr<::mega::MegaUserList> contactList, std::shared_ptr<::mega::MegaTextChatList> chatList)
+{
+    std::unique_ptr<char[]> sid(api.sdk.dumpSession());
```
sergiohs84: Can we please use a regular `const char *` for `sid`?

src/chatClient.h
```diff
@@ -729,6 +729,9 @@ class Client: public rtcModule::IGlobalHandler,
     const char* connStateStr() const { return connStateToStr(mConnState); }
     static const char* connStateToStr(ConnState state);
 
+    void initCacheFromSdkSession(const std::string scsn,
+                                 std::shared_ptr<::mega::MegaUserList> contactList,
```
sergiohs84: Shouldn't we use references to the shared pointers here? analogy as `initWithNewSession()`

src/chatClient.cpp
```diff
@@ -594,23 +594,25 @@ void Client::onRequestFinish(::mega::MegaApi* apiObj, ::mega::MegaRequest *reque
...
-        std::shared_ptr<::mega::MegaTextChatList> chatList(api.sdk.getChatList());
 
         auto wptr = weakHandle();
-        marshallCall([wptr, this, state, scsn, contactList, chatList]()
+        marshallCall([wptr, this]()
         {
             if (wptr.deleted())
                 return;
 
+            char* pscsn = api.sdk.getSequenceNumber();
```
sergiohs84: What's the reason to capture all this data inside the marshal call? despite the action packets are paused, the app sharing the MegaApi instance may do requests in the meantime, which may result on different set of contacts/chats.

tests/sdk_test/sdk_test.cpp
```diff
@@ -648,16 +648,11 @@ bool MegaChatApiTest::TEST_ResumeSession(unsigned int accountIndex)
     megaChatApi[accountIndex]->addChatRequestListener(this);
     megaChatApi[accountIndex]->addChatListener(this);
     MegaChatApi::setLoggerObject(logger);
-    // back to enabled: init + fetchnodes + connect
-    ASSERT_CHAT_TEST(megaChatApi[accountIndex]->init(session) == MegaChatApi::INIT_NO_CACHE,
+    ASSERT_CHAT_TEST(megaChatApi[accountIndex]->enableChat(session),
                      "Wrong chat initialization state. Expected: " + std::to_string(MegaChatApi::INIT_NO_CACHE) + "   Received: " + std::to_string(megaChatApi[accountIndex]->init(session)));
```
sergiohs84: Adjust the messages in case of assertion failed to be consistent with the new call to `enableChat()`

src/chatClient.h
sergiohs84: Based on the review together, not really needed to user references.

src/megachatapi.h
```diff
@@ -1527,11 +1527,6 @@ class MegaChatApi
      * If a session id is provided, karere will try to resume the session from its cache and will
      * return MegaChatApi::INIT_OFFLINE_SESSION.
      *
-     * If a session id is provided but the correspoding cache is not available, it will return
```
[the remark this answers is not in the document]
sergiohs84: In the future, if apps find hard to manage the `INIT_ERROR` due to the library being already initialized, we can add the new state. By now, keep it as is.

src/megachatapi_impl.cpp
sergiohs84: Reject this change, better to let the app know that they are calling `enableChat()` when already initialized

sergiohs84:
A robust approach to correctly manage the initialization of the MEGAchat when the SDK is already up and running is not trivial and will require some changes in the risky zone. Considering the expected usage of this function, very little, we will lower this task (#8688) for the future.

PR #237, opened 2018-03-15
Check cache version during decryption process

alber2510:
Solve problem when cache is removed and there are messages in decryption procedure. 
We have added a cache version that it is incremented when cache is re-created 
At decryption, we check if cache version has changed during procedure


sergiohs84:
This PR is replaced by #240 which covers a missing case (management messages)

PR #242, opened 2018-03-16
Add functionality to notify and receive stop typing

alber2510:


src/chatd.cpp
```diff
@@ -1776,6 +1781,11 @@ Message* Chat::msgModify(Message& msg, const char* newdata, size_t newlen, void*
             return;
         
         postMsgToSending(upd->isSending() ? OP_MSGUPDX : OP_MSGUPD, upd);
+        if (upd->type == Message::kMsgNormal)
```
sergiohs84: I think it's better to not send the notification internally at `msgSubmit()` nor `msgModify()`, since the marshal call could be executed before a previous request to send `isTyping()` is processed, resulting on `isTyping` when it should have been stopped.
The apps need to add the `stopTyping()` anyway for clearing the text box, so we can ask them to do it as well when they send a message.

src/megachatapi.h
```diff
@@ -2617,6 +2619,8 @@ class MegaChatApi
      * @note if MegaChatApi::isMessageReceptionConfirmationActive returns false, messages may never
      * reach the status delivered, since the target user will not send the required acknowledge to the
      * server upon reception.
+     *
+     * Stop typing command is sent automatically
```
sergiohs84: Remove this comment, since the app will take care of stop the notification.

PR #249, opened 2018-03-20
Delay seen messages

jgandres:


src/chatd.cpp
```diff
@@ -1975,42 +1985,52 @@ bool Chat::setMessageSeen(Idx idx)
...
-        if (mLastSeenIdx == CHATD_IDX_INVALID)
         {
-            notifyStart = lownum()-1;
+          return;
         }
-        else
+        mClient.mSeenTimers.erase(mEchoTimer);
+        mEchoTimer = 0;
+        karere::Id id = msg.id();
+        marshallCall([wptr, this, id, idx]()
```
sergiohs84: Theorically, no need the marshal anymore if we have a timer, since the lambda to be executed upon expiration of the timer will happen in the MEGAchat's thread, so the access to the socket is already enforced to be done in the MEGAchat's thread.

src/chatd.cpp
```diff
@@ -1975,42 +1985,52 @@ bool Chat::setMessageSeen(Idx idx)
     }
 
     auto wptr = weakHandle();
-    karere::Id id = msg.id();
-    marshallCall([wptr, this, id, idx]()
+    megaHandle mEchoTimer = karere::setTimeout([this, wptr, idx, &msg, &mEchoTimer]()
```
sergiohs84: Capture the `mEchoTimer` directly, not its memory address.

src/chatd.cpp
```diff
@@ -1975,42 +1985,52 @@ bool Chat::setMessageSeen(Idx idx)
...
-        
-        Idx notifyStart;
-        if (mLastSeenIdx == CHATD_IDX_INVALID)
         {
-            notifyStart = lownum()-1;
+          return;
         }
-        else
+        mClient.mSeenTimers.erase(mEchoTimer);
+        mEchoTimer = 0;
```
sergiohs84: The asignment above is for the pointer, not the value. It doesn't have any effect, so get rid of it.

src/chatd.cpp
```diff
@@ -1975,42 +1985,52 @@ bool Chat::setMessageSeen(Idx idx)
...
-        sendCommand(Command(OP_SEEN) + mChatId + id);
-        
-        Idx notifyStart;
-        if (mLastSeenIdx == CHATD_IDX_INVALID)
         {
-            notifyStart = lownum()-1;
+          return;
         }
-        else
+        mClient.mSeenTimers.erase(mEchoTimer);
```
sergiohs84: You're trying to remove the element matching the memory address, not the value.

src/chatd.cpp
```diff
@@ -1975,42 +1985,52 @@ bool Chat::setMessageSeen(Idx idx)
...
+                if (m.userid != mClient.mUserId)
+                {
+                    CALL_LISTENER(onMessageStatusChange, i, Message::kSeen, m);
+                }
+            }
+            mLastSeenId = id;
+            CALL_DB(setLastSeen, mLastSeenId);
+            CALL_LISTENER(onUnreadChanged);
+        }, mClient.karereClient->appCtx);
+    },200, mClient.karereClient->appCtx);
```
sergiohs84: Better to define a constant for the timeout, something like `kSeenTimeout`

PR #265, opened 2018-04-03
Implement rich links (receive and send)

alber2510:


src/megachatapi_impl.cpp
```diff
@@ -5201,6 +5207,39 @@ MegaChatMessagePrivate::MegaChatMessagePrivate(const MegaChatMessage *msg)
...
+            {
+                const char *text = msg->getRichPreviewText();
+                const char *title = msg->getRichPreviewTitle();
+                const char *description = msg->getRichPreviewDescription();
+                const char *image = msg->getRichPreviewImage();
+                const char *imageFormat = msg->getRichPreviewImageFormat();
+                const char *icon = msg->getRichPreviewIcon();
+                const char *iconFormat = msg->getRichPreviewIconFormat();
+                const char *url = msg->getRichPreviewUrl();
+                MegaChatRichPreview *richPreview = new MegaChatRichPreview(std::string(text), std::string(title), std::string(description),
```
sergiohs84: I think it's better to create a `copy()` for this class and simply call it here, instead of using all this getters, create temporal strings and so on. That's also easier to reuse in the future.

alber2510: I have follow this approach because the the parameter that we receive at this method  is a MegaChatMessage and the interface of this class only has getters to the contained info. We can make a casting to MegaChatMessagePrivate and get the MegaChatRichPreview object instead of get their fields one to one

src/chatd.cpp
```diff
@@ -2346,6 +2510,13 @@ void Chat::rejectMsgupd(Id id, uint8_t serverReason)
         throw std::runtime_error(errorMsg);
     }
 
+    // Update with contains meta has been reject by server. We don't notify
+    if (msg.type == Message::kMsgContainsMeta)
```
sergiohs84: Don't we need here to delete the msg from the sending queue? otherwise, the edit will be retried sooner or later.

src/chatd.cpp
```diff
@@ -1606,6 +1647,117 @@ void Chat::initChat()
...
+        if (!regex_match(url, expresion))
+        {
+            linkRequest = std::string("http://") + url;
+        }
+
+        auto wptr = weakHandle();
+        Idx messageIdx = msgIndexFromId(message.id());
+        uint16_t updated = message.updated;
+        client().karereClient->api.call(&::mega::MegaApi::requestRichPreview, linkRequest.c_str())
+        .then([wptr, this, messageIdx, updated, text](ReqResult result)
```
sergiohs84: I think it's safer to capture the `msgid` rather than the index, since a `REJECT` for `JOINRANGEHIST` may clear the history and the captured index may reference a different message.

src/chatd.cpp
```diff
@@ -1606,6 +1647,117 @@ void Chat::initChat()
...
+                std::string textMessage = result->getText();
+                header.insert(header.begin(), 0x0);
+                header.insert(header.begin(), Message::kMsgContainsMeta);
+                header.insert(header.begin(), 0x0);
+                header = header + std::string("{\"textMessage\":\"") + msg->toText() + std::string("\",\"extra\":[");
+                std::string updateText = header + textMessage + std::string("]}");
+                size_t size = updateText.size();
+
+                msg->type = Message::kMsgContainsMeta;
+                msgModify(*msg, updateText.c_str(), size, NULL);
```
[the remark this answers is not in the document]
sergiohs84: The function was not returning NULL. I added a commit to fix it, but the error should still be managed here.

src/chatd.cpp
```diff
@@ -2391,10 +2564,11 @@ void Chat::onMsgUpdated(Message* cipherMsg)
             it++;
             mPendingEdits.erase(cipherMsg->id());
             mSending.erase(erased);
+            msgUpdatedByMe = true;
```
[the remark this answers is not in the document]
sergiohs84: Fixed by commit 51c1132142f268d367d36ec82e062e24dbdbd420

src/megachatapi_impl.cpp
```diff
@@ -5385,13 +5408,9 @@ MegaChatMessagePrivate::MegaChatMessagePrivate(const Message &msg, Message::Stat
         }
         case MegaChatMessage::TYPE_CONTAINS_META:
         {
-            std::string linkName = JSonUtils::parseContainsMeta(msg.toText().c_str());
-
-            // TODO remove when applications can manage MegaChatMessage::TYPE_CONTAINS_META
-            if (linkName.size())
+            if (msg.toText().length() > 2)
```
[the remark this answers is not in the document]
sergiohs84: OK, during testing I found the reason. The type of meta is 0, which creates issues at `parseContainesMeta()` during parsing of the string. I'll revert the change I made previously :)

PR #268, opened 2018-04-05
Notify to apps webRtc session state

alber2510:


src/megachatapi_impl.cpp
```diff
@@ -5759,21 +5778,46 @@ MegaChatSessionHandler::~MegaChatSessionHandler()
...
+                         Id(chatCall->getId()).toString().c_str(),
+                         session->receivedAv().toString().c_str());
 
-        megaChatApi->fireOnChatCallUpdate(chatCall);
+            chatCall->setSessionStatus(MegaChatCall::SESSION_STATUS_IN_PROGRESS, session->peer());
+
+            megaChatApi->fireOnChatCallUpdate(chatCall);
+            break;
+        }
+        case rtcModule::ISession::kStateDestroyed:
```
sergiohs84: Maybe we should fall-through here for `kStateTerminating`??

src/megachatapi_impl.cpp
```diff
@@ -5759,21 +5778,46 @@ MegaChatSessionHandler::~MegaChatSessionHandler()
...
+    switch (newState)
     {
-        MegaChatCallPrivate* chatCall = callHandler->getMegaChatCall();
-        chatCall->setRemoteAudioVideoFlags(session->receivedAv());
-        API_LOG_INFO("Initial remote audio/video flags. ChatId: %s, callid: %s, AV: %s",
-                     Id(chatCall->getChatid()).toString().c_str(),
-                     Id(chatCall->getId()).toString().c_str(),
-                     session->receivedAv().toString().c_str());
+        case rtcModule::ISession::kStateWaitSdpOffer:
+        case rtcModule::ISession::kStateWaitSdpAnswer:
```
sergiohs84: Should we also add the `kStateWaitLocalSdpAnswer`?

src/megachatapi_impl.cpp
sergiohs84: Not really, better to only notify one of the events. But avoid to notify when the call itself is being terminated.
Done!

src/megachatapi_impl.cpp
sergiohs84: No, that's an intermediate state that doesn't add value to the apps.

PR #280, opened 2018-04-17
Get chatListItems by peers

jgandres:


src/megachatapi_impl.cpp
```diff
@@ -1885,6 +1885,51 @@ MegaChatListItemList *MegaChatApiImpl::getChatListItems()
...
 }
 
+MegaChatListItemList *MegaChatApiImpl::getChatListItemsByPeers(megachat::MegaChatPeerList *peers)
+{
+    MegaChatListItemListPrivate *items = new MegaChatListItemListPrivate();
+
+    sdkMutex.lock();
+
+    if (mClient && !terminating)
+        {
```
sergiohs84: Please, adjust indentation here

src/megachatapi_impl.cpp
```diff
@@ -1885,6 +1885,51 @@ MegaChatListItemList *MegaChatApiImpl::getChatListItems()
...
+            ChatRoomList::iterator it;
+            for (it = mClient->chats->begin(); it != mClient->chats->end(); it++)
+            {
+                bool sameParticipants = true;
+                MegaChatRoomPrivate *chat = new MegaChatRoomPrivate(*it->second);
```
sergiohs84: Instead of creating a new `MegaChatRoomPrivate` just for comparison of the peer list, it would be better to directly use the `it->second` for that purpose. And create the high-level (and kind of more heavy) object only if it's one of the chatrooms to be returned by this function (aka. in the block of `if (sameParticipants == true)`

src/megachatapi_impl.cpp
```diff
@@ -1885,6 +1885,51 @@ MegaChatListItemList *MegaChatApiImpl::getChatListItems()
     return items;
 }
 
+MegaChatListItemList *MegaChatApiImpl::getChatListItemsByPeers(megachat::MegaChatPeerList *peers)
```
sergiohs84: Here we need to remove `megachat` namespace in the parameters. Otherwise, cross-compilers such as Android NDK will complain

src/megachatapi.h
```diff
@@ -2056,6 +2056,25 @@ class MegaChatApi
...
+     * before calling this function.
+     *
+     * Note that MegaChatListItem objects don't include as much information as
+     * MegaChatRoom objects, but a limited set of data that is usually displayed
+     * at the list of chatrooms, like the title of the chat or the unread count.
+     *
+     * This function returns even archived chatrooms.
+     *
+     * You take the ownership of the returned value
+     *
```
sergiohs84: I miss the doc for the `@param peers`. Please, remark the list should not include your own user, since MEGAchat doesn't include them in the map of members for each chatroom.

sergiohs84:
This changes will be incorporated to the `feature/archive-chats`

PR #285, opened 2018-04-19
Group calls

alber2510:


src/megachatapi_impl.cpp
```diff
@@ -5857,17 +6165,18 @@ void MegaChatCallHandler::onStateChange(uint8_t newState)
...
-                             call->chat().chatId().toString().c_str(), call->id().toString().c_str(),
-                              rtcModule::termCodeToStr(call->termCode() & (~rtcModule::TermCode::kPeer)),
+                             karere::Id(chatCall->getChatid()).toString().c_str(),
+                             karere::Id(chatCall->getId()).toString().c_str(),
+                             rtcModule::termCodeToStr(call->termCode() & (~rtcModule::TermCode::kPeer)),
                              chatCall->isLocalTermCode(), chatCall->getDuration());
                 break;
             case rtcModule::ICall::kStateDestroyed:
-                state = MegaChatCall::CALL_STATUS_DESTROYED;
+                return;
```
sergiohs84: Should we use `MegaChatCall::CALL_STATUS_USER_NO_PRESENT` in this case?

src/megachatapi_impl.cpp
```diff
@@ -6077,7 +6454,8 @@ void MegaChatSessionHandler::onRemoteStreamRemoved()
 void MegaChatSessionHandler::onPeerMute(karere::AvFlags av, karere::AvFlags oldAv)
 {
     MegaChatCallPrivate* chatCall = callHandler->getMegaChatCall();
-    chatCall->setRemoteAudioVideoFlags(av);
+    megaChatSession->setAvFlags(session->receivedAv());
```
sergiohs84: Should we use here the received `av` in the params of the callback, rather than the ones from the `session`??

src/megachatapi_impl.cpp
alber2510: I haven't notified `MegaChatCall::CALL_STATUS_USER_NO_PRESENT` at `kStateDestroyed` to avoid notify twice the same state to apps. The call state flow is `in-progress`->`terminating`->`destroyed`. We notify `MegaChatCall::CALL_STATUS_USER_NO_PRESENT` at `kStateTerminating` then I think that it's not necessary notify at `kStateDestroyed` too

src/megachatapi_impl.cpp
alber2510: `av` and `session->receivedAv()` have the same values. It is possible use both

sergiohs84: OK, let's use then the received value in the params for readability.

examples/qtmegachatapi/MainWindow.cpp
```diff
@@ -134,47 +134,97 @@ void MainWindow::onChatCallUpdate(megachat::MegaChatApi */*api*/, megachat::Mega
...
-           }
-           break;
-        case megachat::MegaChatCall::CALL_STATUS_IN_PROGRESS:
+        switch (call->getStatus())
+        {
+            case megachat::MegaChatCall::CALL_STATUS_TERMINATING_USER_PARTICIPATION:
+            {
+                window->hangCall();
+                return;
+                break;
```
jgandres:
```suggestion
```

PR #309, opened 2018-05-10
Chat links

jgandres:
Before deployment of chat-links:
- If account has at least 1 groupchat, the MEGAchat's upgrade v.5 -> v.6 needs to force a full-reload of SDK's cache too.
- If account has no chats, no need to reload SDK's, but simply add the new columns.
 
This PR requires the analog branch from SDK. https://github.com/meganz/sdk/pull/1366


src/megachatapi.h
```diff
@@ -2387,6 +2387,31 @@ class MegaChatApi
      */
     void setChatTitle(MegaChatHandle chatid, const char *title, MegaChatRequestListener *listener = NULL);
 
+    /**
+     * @brief Allows any user to preview an open chat without be part of.
+     *
+     * The associated request type with this request is MegaChatRequest::TYPE_PREVIEW_CHAT_LINK
+     * Valid data in the MegaChatRequest object received on callbacks:
```
sergiohs84: The following 2 lines are incorrect. It's only true when request success, not for all callbacks. I will amend it. The `getText()` should return the link in all callbacks. We should use a different field to return the title if success.

src/megachatapi.h
```diff
@@ -2387,6 +2387,31 @@ class MegaChatApi
...
+     * - MegaChatRequest::getChatHandle - Returns the chat identifier
+     * - MegaChatRequest::getText - Returns the title of the chat.
+     *
+     * On the onRequestFinish error, the error code associated to the MegaChatError can be:
+     * - MegaChatError::ERROR_ARGS - If chatlink has not an appropiate format
+     * - MegaChatError::ERROR_EXIST - If the chat link has been opened before in the current session
```
sergiohs84: `ERROR_EXIST` also in case where the chat is already joined (own user participates). I will change it.

src/megachatapi_impl.cpp
```diff
@@ -692,6 +692,62 @@ void MegaChatApiImpl::sendPendingRequests()
...
+            const char *link = request->getText();
+            string parsedLink = link;
+
+            size_t pos = parsedLink.find("#chat_");
+            if (pos == string::npos)
+            {
+                errorCode = MegaChatError::ERROR_ARGS;
+                break;
+            }
+
```
sergiohs84: Despite it's not a problem to have those numbers hardcoded here, it's good to have a comment to guide the reader. Please add a line with the expected format of the link, something like:
`#chat_<public-handle><chat-key><\0>`

src/megachatapi_impl.cpp
```diff
@@ -692,6 +692,62 @@ void MegaChatApiImpl::sendPendingRequests()
...
+            string key;
+            string keystr = parsedLink.substr(pos);
+            Base64::atob(keystr, key);
+
+            if (ISUNDEF(ph) || key.length() != 16)
+            {
+                errorCode = MegaChatError::ERROR_ARGS;
+                break;
+            }
+
```
sergiohs84: I will add here a check to discard the chatroom is part of the list of chatrooms and/or already in preview mode

src/megachatapi_impl.cpp
```diff
@@ -692,6 +692,62 @@ void MegaChatApiImpl::sendPendingRequests()
...
+            mClient->openChatLink(ph, key)
+            .then([request, this, ph]()
+            {
+                MegaChatHandle openChatId = mClient->chatIdByPh(ph);
+                request->setChatHandle(openChatId);
+
+                auto itChat = mClient->chats->find(openChatId);
```
sergiohs84: We have `MegaChatApiImpl::findChatRoom()` exactly for this purpose :) I will replace this code by a call to such function.

src/chatdDb.h
```diff
@@ -364,6 +364,54 @@ class ChatdSqliteDb: public chatd::DbInterface
...
+        mDb.query(
+            "insert or replace into chat_vars(chatid, name, value) "
+            "values(?, 'open_chat', true)", mChat.chatId());
+        assertAffectedRowCount(1);
+    }
+
+    //Remove open_chat and preview_mode records related to the chatId
+    virtual void removeOpenChat()
+    {
+        mDb.query("delete from chat_vars where chatid = ?", mChat.chatId());
```
sergiohs84: The purpose of this function is to remove records for `open_chat` (and perhaps `preview_mode`) from `chat_vars` table when the chat becomes private, right? If so, you don't want to remove any other variable related to this chat, i.e. the `have_all_history`, which would still be valid for the (now) private chatroom, since we would try to keep the history in the transition from open to closed/private.

src/chatdDb.h
```diff
@@ -364,6 +364,54 @@ class ChatdSqliteDb: public chatd::DbInterface
...
+        assertAffectedRowCount(1);
+    }
+
+    virtual bool openChat()
+    {
+        SqliteStmt stmt(mDb, "select value from history where chatid = ? and name=open_chat");
+        stmt << mChat.chatId();
+        return (stmt.step()) ? stmt.int64Col(0) : false;
```
sergiohs84: Check datatypes for each column in DB schema. Value is not an `int64` but a text string (despite it has the value `1`)
https://github.com/meganz/MEGAchat/blob/develop/src/dbSchema.sql#L26

src/chatdDb.h
```diff
@@ -364,6 +364,54 @@ class ChatdSqliteDb: public chatd::DbInterface
...
+    }
+
+    virtual void setPreviewMode()
+    {
+        mDb.query(
+            "insert or replace into chat_vars(chatid, name, value) "
+            "values(?, 'preview_mode', true)", mChat.chatId());
```
sergiohs84: Instead of `true`, you should use `1`. Take `setHaveAllHistory()` as reference

src/chatdDb.h
```diff
@@ -364,6 +364,54 @@ class ChatdSqliteDb: public chatd::DbInterface
...
+
+    //Cleanup of all records related to the chatId in all DB tables that could contain any record
+    virtual void chatCleanup()
+    {
+        mDb.query("delete from chat_peers where chatid = ?", mChat.chatId());
+        mDb.query("delete from chat_vars where chatid = ?", mChat.chatId());
+        mDb.query("delete from chats where chatid = ?", mChat.chatId());
+        mDb.query("delete from history where chatid = ?", mChat.chatId());
+        mDb.query("delete from manual_sending where chatid = ?", mChat.chatId());
+        mDb.query("delete from sending where chatid = ?", mChat.chatId());
```
sergiohs84: Add `sendkeys` table, just in case :)

src/megachatapi_impl.cpp

src/megachatapi_impl.cpp
sergiohs84: We can actually return requested data if the chat is already opened for preview. And only fail if we already participate in the chatroom as a peer. Already changed that.

src/megachatapi.h
```diff
@@ -981,7 +981,7 @@ class MegaChatRequest
         TYPE_SEND_TYPING_NOTIF, TYPE_SIGNAL_ACTIVITY,
         TYPE_SET_PRESENCE_PERSIST, TYPE_SET_PRESENCE_AUTOAWAY,
         TYPE_LOAD_AUDIO_VIDEO_DEVICES, TYPE_PUSH_RECEIVED,
-        TOTAL_OF_REQUEST_TYPES
```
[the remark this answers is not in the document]
sergiohs84: Done!

sergiohs84:
trigger compilation

src/chatClient.cpp
```diff
@@ -1337,25 +1441,129 @@ void Client::onUsersUpdate(mega::MegaApi* api, mega::MegaUserList *aUsers)
...
+        }
+
+        return createChatPromise
+        .then([this, wptr](ReqResult result) -> Promise<karere::Id>
+        {
+            if (wptr.deleted())
+                return promise::Error("Chat created successfully, but instance was removed");
+
+            auto& list = *result->getMegaTextChatList();
+            if (list.size() < 1)
```
sergiohs84: Actually we can enforce `list.size() != 1` here

src/megachatapi_impl.cpp
```diff
@@ -5287,7 +5287,7 @@ MegaChatRoomPrivate::MegaChatRoomPrivate(const ChatRoom &chat)
     this->chatid = chat.chatid();
     this->priv = (privilege_t) chat.ownPriv();
     this->group = chat.isGroup();
-    this->mPublicChat = ((GroupChatRoom &)chat).publicChat();
+    this->mPublicChat = this->group ?((GroupChatRoom &)chat).publicChat() :false;
```
sergiohs84: Better to use the overloaded method `ChatRoom.publicChat()` directly, which will return `false` for the `PeerChatRoom`.

src/megachatapi_impl.cpp
```diff
@@ -5287,7 +5287,7 @@ MegaChatRoomPrivate::MegaChatRoomPrivate(const ChatRoom &chat)
...
     this->mPreviewMode = ((GroupChatRoom &)chat).previewMode();
```
sergiohs84: Should not cast to `GroupChatRoom` if the `chat` is not groupal.

sergiohs84: It would be good to create `virtual ChatRoom::previewMode() { return false; }`

src/megachatapi_impl.cpp

src/megachatapi_impl.cpp

src/chatd.cpp
```diff
@@ -2505,6 +2505,17 @@ uint64_t Chat::publicHandle()
...
 
+bool Chat::previewMode()
+{
+    return mPh.isValid();
+}
+
+void Chat::rejoin()
+{
+    clearHistory();
+    join();
```
sergiohs84: Do we really need to send a new `JOIN` in this case? I though it would be enough with a `HIST`. If so, we should wrap the same code for `REJECT RANGE 1`:
```
void Chat::refreshHistory()
{
   clearHistory();
   requestHistoryFromServer(-chat.initialHistoryFetchCount);
}
```
And call it in that existing case and also in this new case (avoid the creation of `rejoin()` if so, and call `refreshHistory()` instead).

src/megachatapi.h
```diff
@@ -2709,11 +2709,46 @@ class MegaChatApi
     void inviteToChat(MegaChatHandle chatid, MegaChatHandle uh, int privilege, MegaChatRequestListener *listener = NULL);
 
     /**
-     * @brief TODO documentation
+     * @brief Allow a user to add himself to an existing public chat. To do this the public chat
+     * must have a valid public handle.
```
sergiohs84: Further than to have a valid public handle, I'd say the chatroom must be in preview mode, the result of a previous call to `loadChatLink()` and the public handle contained in the chat-link must be still valid.

src/megachatapi.h
```diff
@@ -2709,11 +2709,46 @@ class MegaChatApi
...
+     * @param listener MegaChatRequestListener to track this request
      */
     void joinChatLink(MegaChatHandle chatid, MegaChatRequestListener *listener = NULL);
 
+    /**
+     * @brief Allow a user to rejoin to an existing public chat. To do this the public chat
+     * must have a valid public handle.
+     *
+     * This function must be called only after calling:
+     *  - MegaChatApi::loadChatLink and receive MegaChatError::ERROR_ACCESS (You are trying to
```
sergiohs84: I'd document this codepath in the `joinChatLink()` interface, which is the place where a developer will check in case they get an error (despite a remarked mention in the documentation here is appreciated too).

src/megachatapi_impl.cpp
```diff
@@ -513,13 +512,25 @@ void MegaChatApiImpl::sendPendingRequests()
 
             if (!chatroom->isGroup()
                 || !chatroom->publicChat()
-                || !chatroom->previewMode())   // Join only for public group chats in preview mode
+                || (!chatroom->previewMode() && chatroom->isActive()))   // Join only for public group chats in preview mode
```
sergiohs84: In preview mode or not in preview mode but inactive (the comment above needs to be updated)

src/megachatapi_impl.cpp
```diff
@@ -793,8 +804,19 @@ void MegaChatApiImpl::sendPendingRequests()
                 request->setChatHandle(chatid);
                 request->setText(room->titleString().c_str());
                 request->setNumber(room->getNumPeers());
+                request->setUserHandle(ph);
+
+                MegaChatErrorPrivate *megaChatError;
+
+                if (room->isActive())
```
sergiohs84: The special condition here is `if (!room->isActive())`, in which case we want to notify the app in a special way. It's better to honor that condition, rather than changing it for the `ok` case. I would write it as:
```
if (!room->isActive())  // a public chat already left cannot be previewed: join it directly.
{
    megaChatError = new MegaChatErrorPrivate(MegaChatError::ERROR_ACCESS);
}
else
{
    megaChatError = new MegaChatErrorPrivate(MegaChatError::ERROR_OK);
}
```

src/megachatapi.h
```diff
@@ -2548,6 +2710,47 @@ class MegaChatApi
...
+     * Valid data in the MegaChatRequest object received on callbacks:
+     * - MegaChatRequest::getChatHandle - Returns the chat identifier
+     * - MegaChatRequest::getUserHandle - Returns the public handle of the chat
+     *
+     * On the onRequestFinish error, the error code associated to the MegaChatError can be:
+     * - MegaChatError::ERROR_ARGS - If the chatid is not valid, the chatroom is not groupal,
+     * the chatroom is not public or the chatroom is in preview mode.
+     * - MegaChatError::ERROR_NOENT - If there isn't any chat with the specified chatid.
+     *
+     * @param chatid MegaChatHandle that identifies the chat room
```
jnavarrom: ph param is missing in the documentation

src/megachatapi.h
```diff
@@ -3842,7 +4148,8 @@ class MegaChatRoom
         CHANGE_TYPE_CLOSED              = 0x20, /// The chatroom has been left by own user
         CHANGE_TYPE_OWN_PRIV            = 0x40, /// Our privilege level has changed
         CHANGE_TYPE_USER_STOP_TYPING    = 0x80, /// User has stopped to typing. \see MegaChatRoom::getUserTyping()
-        CHANGE_TYPE_ARCHIVE             = 0x100 /// Archived or unarchived
+        CHANGE_TYPE_ARCHIVE             = 0X100, /// Archived or unarchived
+        CHANGE_TYPE_CHAT_MODE           = 0x400, /// User has set chat mode to private
```
sergiohs84: Why not `0x200`? Is there another change-type already in use in `develop`, or it's simply a typo due to copy&paste the change-type from the `MegaChatListItem`??

sergiohs84:
> karere-native-linux-PRs FAILED (450) -1

Jenkins does not point to staging, that's the reason for the failure.

src/megachatapi_impl.cpp
jgandres: I think that It could be updated by: Cannot autojoin an active chat if it's not in preview mode.
What do you think?

src/megachatapi_impl.cpp
jgandres: It's deprecated because of rejoin

src/megachatapi.h
jgandres: CHANGE_TYPE_CALL is 0x200

src/chatd.cpp
```diff
@@ -1734,6 +1890,16 @@ bool Chat::isGroup() const
...
 }
 
+bool Chat::isPublic() const
+{
+    return mCrypto->isPublicChat();
+}
+
+uint32_t Chat::getNumPreviewers() const
+{
+    return mNumPreviewers;
```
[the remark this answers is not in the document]
jgandres: Fixed with the following commits:
ae818bd518dab90ced106f219221af0754b96e2d
128a2a39505f762579d339ea03ff458fb658492d

src/chatClient.cpp
jgandres:
```suggestion
            if (list.size() != 1)
```

src/chatClient.cpp
```diff
@@ -411,12 +411,11 @@ promise::Promise<void> Client::sdkLoginExistingSession(const char* sid)
...
 
 
-promise::Promise<ReqResult> Client::openChatPreview(uint64_t publicHandle, const std::string &key)
+promise::Promise<ReqResult> Client::openChatPreview(uint64_t publicHandle)
 {
     auto wptr = weakHandle();
-    std::shared_ptr<std::string> unifiedKey = std::make_shared<std::string>(key);
     return api.call(&::mega::MegaApi::getChatLinkURL, publicHandle)
-    .then([this, unifiedKey, wptr](ReqResult result) -> promise::Promise<ReqResult>
+    .then([this, wptr](ReqResult result) -> promise::Promise<ReqResult>
```
sergiohs84: If the `.then` handler will do nothing, there's no reason to register it. Better to remove it :)

examples/qtmegachatapi/MainWindow.cpp
```diff
@@ -422,10 +422,15 @@ void MainWindow::on_bSettings_clicked()
     auto actPrintMyInfo = menu.addAction(tr("Print my info"));
     connect(actPrintMyInfo, SIGNAL(triggered()), this, SLOT(onPrintMyInfo()));
 
+    menu.addSeparator();
+    auto actRetryPendingConn = menu.addAction(tr("Retry pending connections"));
+    connect(actRetryPendingConn, SIGNAL(triggered()), this, SLOT(onRetryPendingConnections()));
```
sergiohs84: We may want to add 2 options in the menu, rather than this one:
1. Retry pending connections
2. Force reconnect
The first one is the feature already added, the latter is passing a `true` to the `MegaChatApi` function, so it reconnects even if there's a connection already established.

polmr:
trigger compilation

PR #333, opened 2018-06-01
Add pre-support for unknown types of mng messages

sergiohs84:
It allows to decode the unknown type in future attempts once the
MEGAchat adds support.


sergiohs84:
trigger compilation

PR #340, opened 2018-06-06
WIP: #8688. Feature/enable chat function

alber2510:
Task #8688


sergiohs84:
The ticket is closed, since apps will not allow to disable chat.

PR #341, opened 2018-06-06
WIP: #9713. Fast connection upon incoming call

alber2510:


sergiohs84:
Too old PR. Already have a Lean-mode implemented here: https://github.com/meganz/MEGAchat/pull/757

PR #348, opened 2018-06-08
Refactor code - Retry session timer

alber2510:


src/rtcModule/webrtc.cpp
```diff
@@ -823,13 +817,8 @@ void Call::msgCallTerminate(RtMessage& packet)
     {
         EndpointId endpointId(packet.userid, packet.clientid);
         time_t sessionRetryTime = 0;
-        auto itSessionRetyTime = mSessRetriesTime.find(endpointId);
-        if (itSessionRetyTime != mSessRetriesTime.end())
-        {
-            sessionRetryTime = itSessionRetyTime->second;
-        }
-
-        if (sessionRetryTime && (time(NULL) - sessionRetryTime <= RtcModule::kSessSetupTimeout))
```
sergiohs84: Does it mean we don't need the timeout for session-setup any longer?

src/rtcModule/webrtc.cpp
```diff
@@ -734,18 +734,12 @@ void Call::handleMessage(RtMessage& packet)
             TermCode reason = static_cast<TermCode>(data.read<uint8_t>(8));
             chatd::EndpointId endPointId(packet.clientid, packet.userid);
 
-            auto itRetriesTime = mSessRetriesTime.find(endPointId);
-            if (!Session::isTermRetriable(reason) && itRetriesTime != mSessRetriesTime.end())
+            auto itRetries = mSessRetries.find(endPointId);
+            if (!Session::isTermRetriable(reason) && itRetries != mSessRetries.end())
```
sergiohs84: Shouldn't we code it like below for homogeneity?
``` if (!Session::isTermRetriable(reason)
{
    cancelSessionRetryTimer(packet.userid, packet.clientid);
}```

src/rtcModule/webrtc.cpp
alber2510: The timeout is controlled by mSessRetries (std::map<chatd::EndpointId, megaHandle>). If pair (EndpointId, timer handle) exists, there is a session retry in course. When timeout expires, the handle is removed from map

src/rtcModule/webrtc.cpp
alber2510: I've thought but _destroyIfNoSessionsOrRetries(reason)_  has to execute if both conditions are valid, _!Session::isTermRetriable(reason)_ and _itRetries != mSessRetries.end()_
Then I thought that it is more simple doing without using _cancelSessionRetryTimer_
If you considerate opportune I can change it

PR #350, opened 2018-06-11
DNS cache for presenced and chatd

sergiohs84:
In order to speed up the connection establishment to chatd/presenced, this branch adds a in-memory cache for resolved hostnames.

Any reconnection attempt automatically uses cached IPs and perform a DNS lookup in parallel. When the DNS is resolved, if the IPs don't match, it forces a reconnect directly (even if connected).

### DIFF
--- a/src/chatd.cpp
+++ b/src/chatd.cpp
@@ -540,8 +540,8 @@ Promise<void> Connection::reconnect()
                     mLoginPromise.reject(errStr, statusDNS, kErrorTypeGeneric);
                 }
             }
-
-            if (cachedIPs)
+            else if (cachedIPs) // if wsResolveDNS() failed immediately, very likely there's
+            // no network connetion, so it's futile to attempt to connect
             {
                 doConnect();
             }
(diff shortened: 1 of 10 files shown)


src/chatd.cpp
```diff
@@ -506,21 +506,30 @@ Promise<void> Connection::reconnect()
...
+                        && (std::find(ipsv6.begin(), ipsv6.end(), ipv6) != ipsv6.end()));
+
                 if (match)
                 {
                     CHATDS_LOG_DEBUG("DNS resolve matches cached IPs.");
                 }
                 else
                 {
+                    // update DNS cache
+                    bool ret = mDNScache.set(mUrl.host, ipv4, ipv6);
```
jgandres: In this line are you setting the same IP's that you got in `mDNScache.get(mUrl.host, ipv4, ipv6);` ?

jgandres: You should set ipsv4.at(0), isn't it?

PR #353, opened 2018-06-13
Add resolution constraints for group calls

alber2510:


src/rtcModule/webrtc.cpp
```diff
@@ -173,6 +173,40 @@ bool RtcModule::selectDevice(const std::string& devname,
...
+            break;
+
+        case Resolution::low:
+            mMediaConstraints.SetMandatory(webrtc::MediaConstraintsInterface::kMaxHeight, 288);
+            mMediaConstraints.SetMandatory(webrtc::MediaConstraintsInterface::kMinHeight, 240);
+            mMediaConstraints.SetMandatory(webrtc::MediaConstraintsInterface::kMaxWidth, 352);
+            mMediaConstraints.SetMandatory(webrtc::MediaConstraintsInterface::kMinWidth, 320);
+            break;
+
+        case Resolution::vga:
```
sergiohs84: We could use fall-through for this case, since it's the same than for `default`

src/rtcModule/webrtc.cpp
```diff
@@ -336,64 +370,69 @@ void RtcModule::removeCall(Call& call)
...
-        return std::make_shared<artc::LocalStreamHandle>(nullptr, nullptr);
+        catch(exception& e)
+        {
+            audioInput.reset();
+            errors.append("Error getting audio device: ")
+                  .append(e.what()?e.what():"Unknown error")+='\n';
+         }
+     }
+
+   if (!audioInput && !videoInput)
```
sergiohs84: No need for this sanity checkup. `setAv()` is already protected against null audio/video.

src/rtcModule/webrtc.cpp
```diff
@@ -173,6 +173,40 @@ bool RtcModule::selectDevice(const std::string& devname,
...
+            break;
+
+        case Resolution::vga:
+            mMediaConstraints.SetMandatory(webrtc::MediaConstraintsInterface::kMaxHeight, 480);
+            mMediaConstraints.SetMandatory(webrtc::MediaConstraintsInterface::kMinHeight, 480);
+            mMediaConstraints.SetMandatory(webrtc::MediaConstraintsInterface::kMaxWidth, 640);
+            mMediaConstraints.SetMandatory(webrtc::MediaConstraintsInterface::kMinWidth, 640);
+            break;
+
+        default:
```
sergiohs84: Maybe we can add the `notDefined` here?

src/rtcModule/webrtcAdapter.cpp
```diff
@@ -173,7 +174,21 @@ void InputDeviceShared<webrtc::VideoTrackInterface, webrtc::VideoTrackSourceInte
...
-    env->CallStaticVoidMethod(applicationClass, startVideoCaptureMID, (jlong)proxySource.release(), surfaceTextureHelper);
+    if (startVideoCaptureWithParametersMID)
+    {
+        std::string widthString;
+        mOptions->constraints.GetMandatory().FindFirst(webrtc::MediaConstraintsInterface::kMinWidth, &widthString);
+        int width = std::atoi(widthString.c_str());
+        std::string heightString;
+        mOptions->constraints.GetMandatory().FindFirst(webrtc::MediaConstraintsInterface::kMinHeight, &heightString);
+        int height = std::atoi(heightString.c_str());
+        env->CallStaticVoidMethod(applicationClass, startVideoCaptureWithParametersMID, (jint)width, (jint)height, (jint)15, (jlong)proxySource.release(), surfaceTextureHelper);
```
sergiohs84: We could improve readability by declaring `int fps = 15;` here, instead of passing a hardcoded value.

src/rtcModule/webrtc.cpp
alber2510: Solved it

src/rtcModule/webrtc.cpp
alber2510: Solved it

src/rtcModule/webrtcAdapter.cpp
alber2510: Solved it

PR #362, opened 2018-06-25
Hotfix/Improve rich link mangement

alber2510:
- Check if message type is normal before checking if it is a valid rich link
- Improve url parser for rich links
- Skip double quotes before build the json for rich link

### DIFF
--- a/src/chatd.cpp
+++ b/src/chatd.cpp
@@ -1735,24 +1735,28 @@ void Chat::requestRichLink(Message &message)
                 for (std::string::size_type i = 0; i < originalMessage.size(); i++)
                 {
                     char character = originalMessage[i];
-                    if (character == '\n')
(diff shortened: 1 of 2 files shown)


src/chatd.cpp
```diff
@@ -1735,20 +1735,25 @@ void Chat::requestRichLink(Message &message)
...
                         textMessage.push_back('\\');
                         textMessage.push_back('n');
                     }
-                    else
+                    else if (character == '\r')
                     {
                         textMessage.push_back('\\');
                         textMessage.push_back('r');
                     }
+                    else if (character == '"')
```
sergiohs84: We should also escape `‘` (single-quote), `\` (backslash) and all the control characters (ASCII codes from 0 to 31).

PR #369, opened 2018-06-28
Hotfix/last character rich link

alber2510:
Before parsing a string to check if it is a rich link, remove unnecessary last characters ('.', ',', ';', ':', ...)

### DIFF
--- a/src/chatd.cpp
+++ b/src/chatd.cpp
@@ -3791,7 +3791,7 @@ bool Message::hasUrl(const string &text, string &url)
         }
         else
         {
-            if (partialString.size() > 0)
+            if (!partialString.empty())
             {
                 removeUnnecessaryLastCharacters(partialString);
                 if (parseUrl(partialString))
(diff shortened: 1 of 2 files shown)


src/chatd.cpp
```diff
@@ -3860,4 +3852,19 @@ bool Message::parseUrl(const std::string &url)
 
     return regex_match(urlToParse, regularExpresion);
 }
+
+void Message::removeUnnecessaryLastCharacters(string &test)
```
sergiohs84: `test` is not a very suitable name for a parameter. I'd use `buf` or similar.

src/chatd.cpp
```diff
@@ -3860,4 +3852,19 @@ bool Message::parseUrl(const std::string &url)
...
+{
+    char lastCharacter = test[test.size() - 1];
```
sergiohs84: Despite the current usage of this function checks the size is greater than zero beforehand, it seems dangerous to access to the `size() - 1` character without a proper checkup.

PR #372, opened 2018-07-02
Adjust termination code for end call message

alber2510:
Termination code from end call message is translated from one type to other type at UI level.
Conversion:
 - A call with termination code END_CALL_REASON_CANCELLED will be converted to END_CALL_REASON_NO_ANSWER
 - A call with duration > 0 and termination code END_CALL_REASON_ENDED or END_CALL_REASON_FAILED will be converted to END_CALL_REASON_ENDED
 - A call with duration == 0 and termination code END_CALL_REASON_ENDED or END_CALL_REASON_FAILED will be converted to END_CALL_REASON_FAILED

### DIFF
--- a/src/megachatapi_impl.cpp
+++ b/src/megachatapi_impl.cpp
@@ -5650,7 +5650,27 @@ MegaChatMessagePrivate::MegaChatMessagePrivate(const Message &msg, Message::Stat
                 }
 
                 priv = callEndInfo->duration;
-                code = callEndInfo->termCode;
+                switch(callEndInfo->termCode)
+                {
+                    case END_CALL_REASON_CANCELLED:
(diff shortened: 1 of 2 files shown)


src/megachatapi_impl.cpp
```diff
@@ -5650,7 +5650,27 @@ MegaChatMessagePrivate::MegaChatMessagePrivate(const Message &msg, Message::Stat
                 }
 
                 priv = callEndInfo->duration;
-                code = callEndInfo->termCode;
+                switch(callEndInfo->termCode)
+                {
+                    case END_CALL_REASON_CANCELLED:
```
sergiohs84: If this reason for a call termination is obsolete, we should reflect it in the documentation

PR #378, opened 2018-07-06
Allow to edit unconfirmed messages

sergiohs84:
If a message is not confirmed, it should always be editable/deletable,
since the message is never too old


polmr:
trigger compilation

polmr:
trigger compilation

polmr:
trigger compilation

polmr:
trigger compilation

polmr:
trigger compilation

PR #379, opened 2018-07-06
Add functionality to group-calls

alber2510:
- Protocol changes
    - Check if our own user is participating in the call before start to ring 
    - Caller decides when call stops ringing
    - Minor modifications at CALLDATA signalling the call end
    - Improve management of CALLDATA with ringing state
    - Build a MegaChatCall with INCALL command
 - Change at stats
 - Solve issue at call destroy and call try to reconnect session when it wasn't necessary


src/rtcModule/rtcStats.cpp
```diff
@@ -319,8 +336,8 @@ void RtcStats::toJson(std::string& json) const
 {
     json.reserve(10240);
     json ="{";
-    JSON_ADD_STR(cid, mSessionId.toString());
-    JSON_ADD_STR(sid, mSessionId.toString());
+    JSON_ADD_STR(cid, mCallId.toString() + ":" + mSessionId.toString());
+    JSON_ADD_STR(sid, mCallId.toString() + ":" + mSessionId.toString());
```
sergiohs84: Review if sending `sid` is really required or not. Now that `sid` and `cid` values are both added to the `cid`, `sid` should be redundant.

sergiohs84:
trigger compilation

PR #382, opened 2018-07-10
Clean warnings and refactoring

sergiohs84:
- Clean warnings related to unused params.
- Clean warnings related to comparisions between different types: enums
with integers and similar.
- Remove unused methods.
- Refactoring of some members, like `client` --> `mKarereClient` and
similar.
- Fix wrong order of initialization list of members in `ChatClient`.
- Reorder of members in `ChatClient`


polmr:
trigger compilation

PR #383, opened 2018-07-10
 Fix an issue with call cancel termination code 

alber2510:
Furthermore this branch fix test with correct termination code


alber2510:
trigger compilation

PR #387, opened 2018-07-13
Fix race-condition in find last text message

sergiohs84:
It may happen that multiple marshall calls get queued, resulting on more
than one `HIST` sent to server, which is ilegal and causes issues. This
commit aims to protect against parallel requests to fetch messages.

### DIFF
--- a/src/chatd.cpp
+++ b/src/chatd.cpp
@@ -3847,7 +3847,7 @@ bool Chat::findLastTextMsg()
         if (wptr.deleted())
             return;
 
-        if (mOnlineState == kChatStateOnline)
+        if (mOnlineState == kChatStateOnline && !mLastTextMsg.isFetching())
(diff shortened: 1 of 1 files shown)


src/chatd.cpp
```diff
@@ -3847,7 +3847,7 @@ bool Chat::findLastTextMsg()
         if (wptr.deleted())
             return;
 
-        if (mOnlineState == kChatStateOnline)
+        if (mOnlineState == kChatStateOnline && !mLastTextMsg.isFetching())
```
alber2510: I don't know if it would be more correct using `mServerFetchState` in stead of !mLastTextMsg.isFetching() or maybe set the condition inside of `requestHistoryFromServer`, in this way, we avoid send two `HIST` commands in all cases

PR #410, opened 2018-07-24
Fix last-message's timestamp for delete

sergiohs84:
IMPORTANT: this branch is blocked. The behavior of the last-ts is TBD by designers, but apparently the last-activity-ts (used to order the list and to show the last-message ts) should be the one from the newest message that is rendered in the UI (so filters out the deleted messages, ie.).

- Allow last-ts to go backwards in time.
- Do not notify last-ts for every message (needed for load from DB).
- Notify last-ts always that a last-message is notified.


examples/qtmegachatapi/MainWindow.cpp
```diff
@@ -863,7 +895,7 @@ void MainWindow::updateLocalChatListItem(MegaChatListItem *item)
     {
         const megachat::MegaChatListItem *auxItem = itItem->second;
         mLocalChatListItems.erase(itItem);
-        addLocalChatListItem(item);
+        addOrUpdateLocalChatListItem(item);
```
alber2510: It would be possible remove the code of this function `MainWindow::updateLocalChatListItem` and call directly to `addOrUpdateLocalChatListItem(item);`

PR #421, opened 2018-08-10
Modify webRtc stats

alber2510:
Modify webRtc stats
Improve performance to calculate if we should add a new sample

### DIFF
--- a/src/rtcModule/IRtcStats.h
+++ b/src/rtcModule/IRtcStats.h
@@ -35,9 +35,9 @@ struct Sample
             short width = 0;        // width of frame
             short height = 0;       // height of frame
             long bwav = 0;          // bandwidth available
-            long firtx = 0;
-            long plitx = 0;
-            long nacktx = 0;
+            long firtx = 0;         // full intra request
+            long plitx = 0;         // picture loss indication
+            long nacktx = 0;        // Negative Acknowledgement
         } r;    // receive
 
         struct : BwInfo
--- a/src/rtcModule/IRtcStats.h
+++ b/src/rtcModule/IRtcStats.h
@@ -21,6 +21,7 @@ struct Sample
 {
     int64_t ts;
     int lq;         // network quality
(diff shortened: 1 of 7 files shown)


src/rtcModule/IRtcStats.h
```diff
@@ -34,6 +35,9 @@ struct Sample
             short width = 0;        // width of frame
             short height = 0;       // height of frame
             long bwav = 0;          // bandwidth available
+            long firtx = 0;
```
sergiohs84: Please, add some comment to this variable and the ones below

src/rtcModule/rtcStats.cpp
```diff
@@ -351,6 +370,7 @@ void RtcStats::toJson(std::string& json) const
     JSON_SUBOBJ("samples");
         JSON_ADD_SAMPLES(, ts);
         JSON_ADD_SAMPLES(, lq);
+         JSON_ADD_SAMPLES(, f);
```
sergiohs84: It seems there's a spare space?

src/rtcModule/webrtc.cpp
```diff
@@ -2651,7 +2651,7 @@ void Session::submitStats(TermCode termCode, const std::string& errInfo)
     }
 
     std::string stats = mStatRecorder->terminate(info);
-    mCall.mManager.mClient.api.sdk.sendChatStats(stats.c_str());
+    mCall.mManager.mClient.api.sdk.sendChatStats(stats.c_str(), 1380);
```
sergiohs84: It would be better to create a `define` for the port number. What about `CHATSTATS_PORT`?

alber2510: Solved it

src/rtcModule/rtcStats.cpp
alber2510: Solved it

src/rtcModule/IRtcStats.h
alber2510: Solved it

PR #436, opened 2018-09-06
Signal node-attachment messages to chatd

alber2510:
This branch can't be merged until the command OP_NEWNODEMSG will be managed in all shards

### DIFF
--- a/src/chatd.cpp
+++ b/src/chatd.cpp
@@ -4245,8 +4245,8 @@ void FilteredHistory::truncateHistory(Id id)
 {
     if (id.isValid())
     {
-        auto it = mIndexMap.find(id);
-        if (it != mIndexMap.end())
+        auto it = mIdToMsgMap.find(id);
+        if (it != mIdToMsgMap.end())
         {
             // id is a message in the history, we want to remove from the next message until the oldest
             for (auto loopIterator = ++it->second; loopIterator != mBuffer.end(); loopIterator++)
--- a/src/chatd.cpp
+++ b/src/chatd.cpp
@@ -3212,6 +3212,7 @@ void Chat::handleTruncate(const Message& msg, Idx idx)
 
 
     Id nextAttachmentId = Id::inval();
+    // lownum is the message where the history has been truncate
     for (Idx i = lownum(); i < highnum(); i++)
     {
         if (at(i).type == Message::kMsgAttachment)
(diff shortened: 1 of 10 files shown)


src/chatd.cpp
```diff
@@ -4144,4 +4178,97 @@ bool Message::isValidEmail(const string &buf)
...
+    Message* message = new Message(msg);
+    if (isNew)
+    {
+        mBuffer.emplace_front(message);
+        mNewest ++;
+        mDb->addAttachmentMessageToHistoryNode(msg, mNewest);
+    }
+    else
+    {
+        if (find(msg.id()) == mBuffer.end())
```
sergiohs84: This operation (insert old attachments retrieved from server) may become too heavy with a long history, since for every message we need to go through the whole list. Would it be possible to assume the message is not part of the node-history yet and add it always?

src/chatd.cpp
```diff
@@ -4144,4 +4178,97 @@ bool Message::isValidEmail(const string &buf)
...
+        }
+    }
+}
+
+void FilteredHistory::deleteMessage(const Message &msg)
+{
+    auto it = find(msg.id());
+    if (it != mBuffer.end())
+    {
+        it->reset(new Message(msg));
```
sergiohs84: Can we simply update the content and the `ts` of the message rather than create a new one and destroy the older msg?

src/chatdDb.h
```diff
@@ -413,6 +428,42 @@ class ChatdSqliteDb: public chatd::DbInterface
...
+
+    virtual void removeAttachmentMessageToHistoryNode(const chatd::Message& msg)
+    {
+        mDb.query("update node_history set type = ?, data = ?, updated = ?, userid = ?, is_encrypted = ? where chatid = ? and msgid = ?",
+                  msg.type, msg, msg.updated, msg.userid, msg.isEncrypted(), mChat.chatId(), msg.id());
+    }
+
+    virtual void truncateHistoryNode(karere::Id id)
+    {
+        auto idx = getIdxOfMsgid(id, "node_history");
```
sergiohs84: If the `id` of the truncate-message is not part of node-history, the returned index will be `CHAT_INVALID_IDX`, which is higher than any valid index, resulting on the full node-history cleared up (which is not the expected behavior).

src/chatd.cpp
alber2510: Solved it

src/chatd.cpp
alber2510: Solved it

src/chatd.cpp
```diff
@@ -2019,7 +2022,14 @@ void Chat::msgSubmit(Message* msg, SetOfIds recipients)
     }
     onMsgTimestamp(msg->ts);
 
-    postMsgToSending(OP_NEWMSG, msg, recipients);
+    if (msg->size() > 2 && msg->buf()[0] == 0 && msg->type == Message::Type::kMsgAttachment)
```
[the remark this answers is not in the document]
alber2510: Yes, not necessary

sergiohs84:
trigger compilation

sergiohs84:
trigger compilation

PR #444, opened 2018-09-14
New termination code to notify when a call is finished by a collision

alber2510:


src/rtcModule/webrtc.h
```diff
@@ -96,6 +96,7 @@ enum TermCode: uint8_t
     kCallGone = 8,
     kBusy = 9,                  // < Peer is in another call
     kNotFinished = 10,          // < It is no finished value, it is TermCode value while call is in progress
+    kDestroyByCallCollision = 11,// < The call has finished by a call collision
```
sergiohs84: In case of really need a new termcode, I'd like to avoid any future collision with upcoming termcodes (in case the protocol eventually needs them).

alber2510: Yes, it's right. I modify the termCode value

PR #446, opened 2018-09-24
Node history - phase 2

alber2510:
- Allow to fetch messages with attachment nodes
 - Notify to apps new/old messages from node history
 - Implement at Qt app node history


src/chatd.cpp
```diff
@@ -4315,11 +4471,116 @@ void FilteredHistory::clear()
...
+
+    uint32_t messagesLoadedFromdb = loadHistoryFromDb(count);
+    mOldestNotifyMsg = --mBuffer.end();
+    if (messagesLoadedFromdb)
+    {
+        mFirstNotification = false;
+        return HistSource::kHistSourceDb;
+    }
+
+    if (!hasAllHistory() && mChat->connection().isOnline())
```
sergiohs84: Here you probably want to check that the connection is actually logged-in into chatd, not just the connection is opened. Note that the connection may be online, but the chat may not be logged-in yet.
I'll change that. Let me know if my assumption is wrong.

src/chatd.cpp
```diff
@@ -4315,11 +4471,116 @@ void FilteredHistory::clear()
...
+    {
+        mFetchingFromServer = true;
+        mChat->requestNodeHistoryFromServer(count);
```
sergiohs84: This line breaks the abstraction principle of the `FilteredHistory` class. It was designed to support different types of history, which would be very useful in the future to load different kind of messages, rather than node-attachments. If we want to achieve that goal, we should somehow instantiate objects of `FilteredHistory` by specifying the type of filter.
Let's discuss about it tomorrow in the office :)

src/chatd.cpp
```diff
@@ -4315,11 +4471,116 @@ void FilteredHistory::clear()
     init();
 }
 
+HistSource FilteredHistory::loadHistory(uint32_t count)
+{
+    uint32_t messagesLoaded = 0;
+    if ((mOldestNotifyMsg != --mBuffer.end() && mBuffer.size() > 1) || mFirstNotification)
```
sergiohs84: I don't understand this checkup. First of all, we should notify messages from RAM if there are messages after `mOldestNotifyMsg`, even if there's only one message, right?
Second, the `mFirstNotification` seems to be useless, since it's reset in all the code-paths here. As long as there are messages available and not-yet-notified in RAM, we should go for that `HistSource`.
Please, let me know if I'm wrong.

PR #449, opened 2018-09-26
Public interface to check if logged in into all chats

sergiohs84:


src/megachatapi.h
```diff
@@ -1941,31 +1938,47 @@ class MegaChatApi
...
-     * @brief Returns the current state of the connection to chatd
+     * @brief Returns the current state of the connection to chatd for a given chatroom
      *
      * The possible values are:
      *  - MegaChatApi::CHAT_CONNECTION_OFFLINE      = 0
      *  - MegaChatApi::CHAT_CONNECTION_IN_PROGRESS  = 1
      *  - MegaChatApi::CHAT_CONNECTION_LOGGING      = 2
      *  - MegaChatApi::CHAT_CONNECTION_ONLINE       = 3
      *
+     * The connection state of all chats can be checked by passing MEGACHAT_INVALID_HANDLE
```
alber2510: I think getChatConnectionState returns MegaChatApi::CHAT_CONNECTION_OFFLINE if MEGACHAT_INVALID_HANDLE is passed as chatid

sergiohs84: You're correct. It was a leftover from the initial approach.
Fixed.

PR #458, opened 2018-10-18
Improve behaviour for webrtc session collision

alber2510:


src/rtcModule/webrtc.cpp
```diff
@@ -2690,21 +2734,20 @@ void Session::submitStats(TermCode termCode, const std::string& errInfo)
...
     {
-        match &= (hash.data[i] == peerHash.data[i]);
+        match &= (hash.data[i] == mPeerHash.data[i]);
     }
     return match;
 }
 
 void Session::msgIceCandidate(RtMessage& packet)
 {
-    assert(!mPeerSdp.empty());
```
sergiohs84: Perhaps it would be good to keep it in this way:
```assert(!mPeerSdpAnswer.empty() || !mPeerSdpOffer.empty());```

PR #469, opened 2018-11-02
Do not update the `mNextHistFetchIdx` upon first login

sergiohs84:
If app tries to load messages before first join and there's no local
history available yet, they received a `HistSource ==
kSourceNotLoggedIn`. During login, received messages won't be notified,
but after login the app can attempt to load messages again and should be
notified about messages from the beginning, not from the last received
message.


sergiohs84:
This commit was included in the https://github.com/meganz/MEGAchat/pull/467

PR #498, opened 2018-11-19
Implement last changes at groupCalls

alber2510:
- New CALLDATA state
- Hangup the call when you are removed from chatroom
- New termination code kErrIceTimeout
- Other minor changes

### DIFF
--- a/src/chatd.cpp
+++ b/src/chatd.cpp
@@ -1058,7 +1058,6 @@ HistSource Chat::getHistoryFromDbOrServer(unsigned count)
     }
     else //have to fetch history from server
     {
-        mServerOldHistCbEnabled = true;
         if (mHaveAllHistory)
         {
             CHATID_LOG_DEBUG("getHistoryFromDbOrServer: No more history exists");
@@ -1083,6 +1082,7 @@ HistSource Chat::getHistoryFromDbOrServer(unsigned count)
                 requestHistoryFromServer(-count);
             }, mClient.karereClient->appCtx);
         }
+        mServerOldHistCbEnabled = true;
         return kHistSourceServer;
     }
 }
@@ -1376,7 +1376,9 @@ void Connection::execCommand(const StaticBuffer& buf)
                 else if (op == OP_RANGE && reason == 1)
                 {
                     chat.clearHistory();
-                    chat.requestHistoryFromServer(-chat.initialHistoryFetchCount);
+                    // we were notifying NEWMSGs in result of JOINRANGEHIST, but after reload we start receiving OLDMSGs
+                    chat.mServerOldHistCbEnabled = mChatdClient.karereClient->isChatRoomOpened(chatid);
+                    chat.getHistoryFromDbOrServer(chat.initialHistoryFetchCount);
                 }
                 else if (op == OP_NEWKEY)
                 {
@@ -1633,7 +1635,14 @@ void Chat::onFetchHistDone()
         else
         {
             mServerFetchState = kHistNotFetching;
-            mNextHistFetchIdx = lownum()-1;
+            // if app tries to load messages before first join and there's no local history available yet,
+            // they received a `HistSource == kSourceNotLoggedIn`. During login, received messages won't be
+            // notified, but after login the app can attempt to load messages again and should be notified
+            // about messages from the beginning
+            if (!mIsFirstJoin)
+            {
+                mNextHistFetchIdx = lownum()-1;
+            }
         }
         if (mLastServerHistFetchCount <= 0)
         {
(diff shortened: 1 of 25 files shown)


src/rtcModule/webrtc.cpp
```diff
@@ -1811,6 +1893,25 @@ bool Call::cancelSessionRetryTimer(karere::Id userid, uint32_t clientid)
     return false;
 }
 
+void Call::monitorCallSetupTimeout()
+{
+    if (!mCallSetupTimer)
```
sergiohs84: Is this checkup really needed?

src/rtcModule/webrtc.cpp
```diff
@@ -2192,11 +2328,18 @@ Session::Session(Call& call, RtMessage& packet, SdpKey sdpkey)
     mName = "sess[" + mSid.toString() + "]";
     auto wptr = weakHandle();
     mSetupTimer = setTimeout([wptr, this] {
-        if (wptr.deleted())
+        if (wptr.deleted() || mState >= kStateInProgress)
+        {
```
sergiohs84: In case the state is in-progress or higher, shouldn't this timer have been stopped already?

src/rtcModule/webrtc.cpp
```diff
@@ -2192,11 +2328,18 @@ Session::Session(Call& call, RtMessage& packet, SdpKey sdpkey)
     mName = "sess[" + mSid.toString() + "]";
     auto wptr = weakHandle();
     mSetupTimer = setTimeout([wptr, this] {
-        if (wptr.deleted())
+        if (wptr.deleted() || mState >= kStateInProgress)
```
sergiohs84: To avoid problems, this timer should be cancelled if the session is setup correctly, doesn't it?
Actually, that's done for the timer that monitors the call-setup. Why not for the session-setup?

src/rtcModule/webrtc.cpp
alber2510: In actual code state, this timer is not stopped. we can change for that behaviour

src/rtcModule/webrtc.cpp
alber2510: No, it's not necessary

src/megachatapi_impl.cpp
```diff
@@ -6874,12 +6907,6 @@ void MegaChatCallHandler::addParticipant(Id userid, uint32_t clientid, AvFlags f
     if (chatCall)
     {
         bool notify = chatCall->addOrUpdateParticipant(userid, clientid, flags);
-
-        if (chatCall->getNumParticipants() == 2 && !chatCall->getInitialTimeStamp())
```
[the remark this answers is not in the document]
alber2510: Because addParticipant is call when we receive a INCALL command. INCALL command can arrive earlier than some session is established

src/megachatapi_impl.cpp
```diff
@@ -1018,7 +1018,7 @@ void MegaChatApiImpl::sendPendingRequests()
             }
 
             MegaChatCallHandler *handler = findChatCallHandler(chatid);
-            if (handler && handler->getCall())
+            if (handler && !chatroom->isGroup())
```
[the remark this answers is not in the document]
alber2510: Yes, we don't want to allow. But start request allow start a call and join to a group call. With this code we want to avoid that an user start a call in an 1to1 chatroom where there is a call and  we allow that a user can join in a groupcall

src/chatd.cpp
```diff
@@ -3809,7 +3842,21 @@ void Chat::onUserLeave(Id userid)
...
+        mClient.mRtcHandler->onKickedFromChatRoom(mChatId);
+#endif
     }
+#ifndef KARERE_DISABLE_WEBRTC
+    else
+    {
+        uint32_t clientid = 0;
+        if (mClient.mRtcHandler->getClientidFromPeer(mChatId, userid, clientid))
+        {
+            onEndCall(userid, clientid);
```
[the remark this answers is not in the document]
alber2510: It's a double check, in case the client of the user, that has been removed from chat room, doesn't finish its call participation, we finish our session with it. With our code, this case isn't possible but with some client modification would be possible

src/chatd.cpp
```diff
@@ -3809,7 +3842,21 @@ void Chat::onUserLeave(Id userid)
     if (userid == client().userId())
     {
         mOwnPrivilege = PRIV_NOTPRESENT;
+#ifndef KARERE_DISABLE_WEBRTC
+        mClient.mRtcHandler->onKickedFromChatRoom(mChatId);
```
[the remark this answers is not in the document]
alber2510: No, I will add the check

src/rtcModule/webrtc.cpp
sergiohs84: Cleaned already

src/megachatapi_impl.cpp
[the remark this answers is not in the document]
sergiohs84: Fixed: avoid a second call to be started in the same groupchat

PR #503, opened 2018-11-26
Promise fix

amdjachenko:
The overall outcome is timer/promise facilities are improved in term of thread safety and UB + many handcrafted replacements for std means are removed. Promises are almost completely reimplemented keeping the same interface and behavior. Not sure that new promise implementation is significantly clearer than the original but that is my best efforts.


sergiohs84:
trigger compilation

amdjachenko:
compilation is fixed, please continue review

sergiohs84:
This PR accumulates too many changes. It would require deep testing just for the shake of a refactoring. The relevant fix will be merged with the PR https://github.com/meganz/MEGAchat/pull/550

PR #506, opened 2018-11-27
Fix issues at group calls

alber2510:


src/megachatapi_impl.cpp
```diff
@@ -1232,7 +1234,16 @@ void MegaChatApiImpl::sendPendingRequests()
                 break;
             }
 
-            if (chatCall->adjustAvFlagsToRestriction(newFlags))
+            ChatRoom *chatroom = findChatRoom(chatid);
+            if (!chatroom)
```
sergiohs84: Is it even possible? I don't think so, but not sure

src/megachatapi_impl.cpp
```diff
@@ -1068,7 +1068,8 @@ void MegaChatApiImpl::sendPendingRequests()
 
                 MegaChatCallPrivate *chatCall = handler->getMegaChatCall();
                 assert(chatCall);
-                if (chatCall->adjustAvFlagsToRestriction(avFlags))
+                EndpointId endPoint(chatroom->chat().client().myHandle(), chatroom->chat().connection().clientId());
+                if (chatCall->adjustAvFlagsToRestriction(avFlags, endPoint))
```
sergiohs84: The logic in `adjustAvFlagsToRestriction()` needs to be reviewed, @alber2510 
It's not doing what you expect and simply updating the video-flag if *any* flag is changed is not the expected behavior. Here and in the 3 other usages of that method.

src/rtcModule/webrtc.cpp
```diff
@@ -645,6 +645,24 @@ bool RtcModule::isCallInProgress(Id chatid) const
     return callInProgress;
 }
 
+bool RtcModule::isCallActive(Id chatid) const
+{
+    if (chatid.isValid())
+    {
+        auto it = mCallHandlers.find(chatid);
```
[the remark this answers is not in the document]
sergiohs84: Done!

PR #508, opened 2018-11-27
Allow several clients from same user in a call

alber2510:


src/megachatapi.h
```diff
@@ -471,18 +497,47 @@ class MegaChatCall
...
+
     /**
      * @brief Get a list with the ids of peers that are participating in the call
      *
      * In a group call, this function returns the list of active participants,
      * regardless your own user participates or not. In consequence,
-     * the list can differ from the one returned by MegaChatCall::getSessions
+     * the list can differ from the one returned by MegaChatCall::getSessionsPeerid
+     *
+     * To idendentify completely a call participant it's necessary the peerid plus the clientid
```
```suggestion
     * To identify completely a call participant it's necessary the peerid plus the clientid
```

src/megachatapi.h
```diff
@@ -2586,6 +2641,18 @@ class MegaChatApi
...
+     * @brief Returns the client id handle of the logged in user for a chatroom
+     *
+     * The clientid is not the same for all chatrooms. If \c chatid is invalid, this function
+     * returns 0
+     *
+     * In offline mode (MegaChatApi::INIT_OFFLINE_SESSION), this function returns 0
+     *
+     * @return Own client id handle
+     */
+    MegaChatHandle getMyClientidHandle(MegaChatHandle chatid);
```
sergiohs84: This method should be at `MegaChatApi` level, rather than in `MegaChatCall`

src/megachatapi.h
```diff
@@ -440,26 +447,45 @@ class MegaChatCall
     /**
      * @brief Get a list with the ids of peers that have a session with me
      *
+     * It's necessary to call MegaChatCall::getSessionsClientid to know the clientid. With clientid and
+     * sessionid, you can identify a Session. We can have a session with the same peerid but with different clientids
```
sergiohs84: I would add some details:
```Every session is identified by a pair of \c peerid and \c clientid. This method returns the list of peerids for each session. Note that, if there are multiple sessions with the same peer (who uses multiple clients), the same peerid will be included multiple times (once per session).```

File: src/rtcModule/webrtc.cpp [hunk 4]
```diff
@@ -1292,6 +1298,9 @@ void Call::msgJoin(RtMessage& packet)
 {
     if (mState == kStateRingIn && packet.userid == mManager.mKarereClient.myHandle())
     {
+        // Avoid new CALLDATA, from caller, starts a new ringing call
+        // Ohter client from our user has answer the call
```
sergiohs84:
```suggestion
        // (another client from our user has already answered the call)
```

src/rtcModule/webrtc.cpp
```diff
@@ -1292,6 +1298,9 @@ void Call::msgJoin(RtMessage& packet)
 {
     if (mState == kStateRingIn && packet.userid == mManager.mKarereClient.myHandle())
     {
+        // Avoid new CALLDATA, from caller, starts a new ringing call
```
sergiohs84:
```suggestion
        // Another client of our own user has already answered the call
        // --> add the participant to the call, so when the upcoming CALLDATA from caller
        // (indicating that a session has been established with our user, but other users should
        // keep ringing), this client does not start ringing again
```

PR #522, opened 2018-12-04
Call RTCMD messages at hangup call instead of destroy

alber2510:


alber2510:
Change has been implemted at webClient. We can close the PR

PR #551, opened 2019-01-09
Add function to retrieve messages from node history

alber2510:


src/megachatapi.h
```diff
@@ -3109,6 +3108,19 @@ class MegaChatApi
...
+     *
+     * This function allows to retrieve only those messages that are in the node history
+     *
+     * You take the ownership of the returned value.
+     *
+     * @param chatid MegaChatHandle that identifies the chat room
+     * @param msgid MegaChatHandle that identifies the message (msg id or a temporal id)
+     * @return The MegaChatMessage object, or NULL if not found.
+     */
+    MegaChatMessage *getMessageFromNodeHistroy(MegaChatHandle chatid, MegaChatHandle msgid);
```
sergiohs84: There's a typo requiring refactoring: `getMessageFromNodeHistroy()` --> `getMessageFromNodeHistory()`

src/megachatapi.h
```diff
@@ -3109,6 +3108,19 @@ class MegaChatApi
...
 
+    /**
+     * @brief Returns the MegaChatMessage specified from the chat room stored in node history
+     *
+     * This function allows to retrieve only those messages that are in the node history
+     *
+     * You take the ownership of the returned value.
+     *
+     * @param chatid MegaChatHandle that identifies the chat room
+     * @param msgid MegaChatHandle that identifies the message (msg id or a temporal id)
```
sergiohs84: A message in the node-history buffer is always confirmed. When they are in sending status they are not included in this buffer, only upon confirmation
```suggestion
     * @param msgid MegaChatHandle that identifies the message
```

src/megachatapi.h
alber2510: Ok, I also think that it's a clearer solution

PR #563, opened 2019-01-15
Consider timestamp of lastest message as last seen

jgandres:
- Create a Map of userids (key) and last green (value)
- Update last seen with last message ts if it's more recent
- Avoid update last seen with last message ts if we have requested last
seen to server and we don't have received any response

### DIFF
--- a/src/chatClient.cpp
+++ b/src/chatClient.cpp
@@ -1149,10 +1149,15 @@ void Client::onPresenceLastGreenUpdated(Id userid, uint16_t lastGreen)
     time_t lastGreenTs = time(NULL) - (lastGreen * 60);
 
     // Update and notify last green if needed
-    updateAndNotifyLastGreen(userid.val, lastGreenTs, true);
+    updateAndNotifyLastGreen(userid.val, lastGreenTs);
 }
 
-void Client::updateAndNotifyLastGreen(Id userid, time_t lastGreenTs, bool force)
+time_t Client::getLastGreen(Id userid)
+{
+    return mPresencedClient.getLastGreen(userid.val);
+}
+
+void Client::updateAndNotifyLastGreen(Id userid, time_t lastGreenTs)
 {
     // Get latest message ts from cache
     time_t lastMsgTs = 0;
@@ -1167,7 +1172,7 @@ void Client::updateAndNotifyLastGreen(Id userid, time_t lastGreenTs, bool force)
     time_t auxLastGreen = (lastGreenTs >= lastMsgTs) ? lastGreenTs : lastMsgTs;
 
     // Update last green if required and notify apps
-    if (mPresencedClient.updateLastGreen(userid.val, auxLastGreen, force))
+    if (mPresencedClient.updateLastGreen(userid.val, auxLastGreen))
     {
         // Format ts to minutes
         time_t auxTs = (time(NULL));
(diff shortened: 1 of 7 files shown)


src/presenced.cpp
```diff
@@ -1272,6 +1307,13 @@ void Client::removePeer(karere::Id peer, bool force)
 
     mCurrentPeers.erase(it);
 
+    // Remove peer from mPeersLastGreen map if exists
+    auto itAux = mPeersLastGreen.find(peer.val);
```
sergiohs84: You can directly do `mPeersLastGreen.erase(peer.val);`, it is equivalent and shorter :)

src/presenced.cpp
```diff
@@ -166,9 +166,44 @@ bool Client::setLastGreenVisible(bool enable)
...
 bool Client::requestLastGreen(Id userid)
 {
+    // Avoid send OP_LASTGREEN if user is ex-contact
+    if (isExContact(userid))
+    {
+        return false;
+    }
+
+    // Reset user last green or insert an entry in the map if not exists
+    std::map<uint64_t, time_t>::iterator it;
```
sergiohs84: All these lines can be simply replaced by `mPeersLastGreen[userid.val] = 0;`. Simplier and shorter :)

src/chatd.cpp
```diff
@@ -1540,6 +1540,8 @@ void Connection::execCommand(const StaticBuffer& buf)
                 }
                 else
                 {
+                    mChatdClient.mKarereClient->updateAndNotifyLastGreen(msg->userid, msg->ts, false);
```
sergiohs84: We don't want to execute this line for every `OLDMSG`, it should be only called `if (opcode == OP_NEWMSG))`

sergiohs84: In any case, probably it's better to call this method once the message is decrypted. Some management messages initially have the `API_USER`/`COMMANDER` as the `userid` and after parsing they change to the user who performed the action (update privileges, change title...) and we would like to update the timestamp in those cases too, if appropriate.

src/chatClient.cpp
```diff
@@ -1145,7 +1145,36 @@ void Client::onPresenceConfigChanged(const presenced::Config& state, bool pendin
 
 void Client::onPresenceLastGreenUpdated(Id userid, uint16_t lastGreen)
 {
-    app.onPresenceLastGreenUpdated(userid, lastGreen);
+    // Calculate lastGreen ts from the system time and format
+    time_t lastGreenTs = time(NULL);
+    lastGreenTs -= (lastGreen * 60);
```
sergiohs84: I think `time_t lastGreenTs = time(NULL) - (lastGreen * 60);` improves readability

src/chatClient.cpp
```diff
@@ -1145,7 +1145,36 @@ void Client::onPresenceConfigChanged(const presenced::Config& state, bool pendin
...
+    time_t lastMsgTs = 0;
+    SqliteStmt stmt(db, "select MAX(ts) from history where userid = ?");
+    stmt << userid.val;
+    if (stmt.step())
+    {
+        lastMsgTs = stmt.uintCol(0);
+    }
+
+    // Compare lastGreen with last message ts and get the bigger
+    time_t auxLastGreen = (lastGreenTs >= lastMsgTs) ?lastGreenTs :lastMsgTs;
```
sergiohs84:
```suggestion
    time_t auxLastGreen = (lastGreenTs >= lastMsgTs) ? lastGreenTs : lastMsgTs;
```

polmr:
trigger compilation

PR #566, opened 2019-01-17
WIP:  Attach video after unmute if video is not defined 

alber2510:


alber2510:
It's not possible to add a video stream after session has been established. When audio/video stream can be removed from a session, we have to stduy how implement this functionality

PR #570, opened 2019-01-22
Added command apiurl to megaclc

mattw-mega:
Also addressed gettimeofday() usage under MSVC

### DIFF
--- a/examples/megaclc/megaclc.cpp
+++ b/examples/megaclc/megaclc.cpp
@@ -1606,6 +1606,56 @@ void exec_getchatcallsids(ac::ACState&)
 
 #endif
 
+class OneShotRequestListener : public m::MegaRequestListener
+{
+public:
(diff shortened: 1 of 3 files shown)


examples/megaclc/megaclc.cpp
```diff
@@ -1606,6 +1606,56 @@ void exec_getchatcallsids(ac::ACState&)
...
+    void onRequestTemporaryError(m::MegaApi *api, m::MegaRequest *request, m::MegaError* error) override
+    {
+        if (onRequestTemporaryErrorFunc) onRequestTemporaryErrorFunc(api, request, error);
+    }
+};
+
+
+void exec_apiurl(ac::ACState& s)
+{
+    if (g_megaApi->isLoggedIn())
```
sergiohs84: The API_URL can be changed even once logged in. All you need to do, in order to keep consistency of data in cache, is a new `fetchnodes()`, so the current cache is discarded and a new fresh one is created from server's data

PR #571, opened 2019-01-23
Add functions to return max call participants and max video participants

alber2510:


src/megachatapi.h
```diff
@@ -3984,6 +3984,20 @@ class MegaChatApi
      */
     bool areGroupChatCallEnabled();
 
+    /**
+     * @brief Returns the maximum call participants
+     *
+     * @return Maximun call participants
```
```suggestion
     * @return Maximum call participants
```

src/megachatapi.h
```diff
@@ -3984,6 +3984,20 @@ class MegaChatApi
...
+     */
+    int getMaxCallParticipants();
+
+    /**
+     * @brief Returns the maximum video call participants
+     *
+     * @return Maximun video call participants
```
```suggestion
     * @return Maximum video call participants
```

PR #578, opened 2019-01-28
Don't allow unaligned memory access for arm64 devices

jnavarrom:
The Undefined Behavior Sanitizer (UBSan) detects a Misaligned Pointer / Reference. 

screenshot 2019-01-28 at 11 18 07

A misaligned pointer has undefined behavior, and may result in a crash or degraded performance.

The problem is that `__arm__` is defined for 32bit arm, and 32bit arm only. At least. when using `clang`.

This pull request tries to solve this problem for arm64 devices.

### DIFF
--- a/src/buffer.h
+++ b/src/buffer.h
@@ -6,7 +6,7 @@
 #include <string.h>
 #include <vector>
 
-#if !defined(__arm__) && !defined(__aarch64__)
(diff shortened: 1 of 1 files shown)


polmr:
trigger compilation

polmr:
trigger compilation

sergiohs84:
trigger compilation

PR #579, opened 2019-01-28
Support to refresh the chatd/presenced URLs and reconnect

sergiohs84:
When API path is changed to staging, not only a full fetchnodes is required, but also the URL for presenced is different, so it needs to be re-fetched and reconnected.

### DIFF
--- a/src/chatd.cpp
+++ b/src/chatd.cpp
@@ -363,15 +363,7 @@ void Chat::connect()
                 return;
             }
 
-            const char* url = result->getLink();
-            if (!url || !url[0])
-            {
-                CHATID_LOG_ERROR("No chatd URL received from API");
-                return;
-            }
-
-            std::string sUrl = url;
-            mConnection.mUrl.parse(sUrl);
+            mConnection.mUrl.parse(result->getLink());
             mConnection.mUrl.path.append("/").append(std::to_string(Client::chatdVersion));
 
             mConnection.reconnect()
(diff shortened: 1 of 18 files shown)


sergiohs84:
trigger compilation

src/chatd.cpp
```diff
@@ -876,36 +868,69 @@ void Connection::doConnect()
...
+    }
 
-            setState(kStateDisconnected);
-            abortRetryController();
-            reconnect();
-        }
-        else if (mRetryCtrl && mRetryCtrl->state() == rh::State::kStateRetryWait)
+    if (refreshURL || !mUrl.isValid())
+    {
+        if (State::kStateFetchingUrl)
```
jgandres:
```suggestion
        if (mState == State::kStateFetchingUrl)
```

src/chatd.cpp
```diff
@@ -876,36 +868,69 @@ void Connection::doConnect()
     }
 }
 
-void Connection::retryPendingConnection(bool disconnect)
+void Connection::retryPendingConnection(bool disconnect, bool refreshURL)
 {
-    if (mUrl.isValid())
+    if (mState == kStateNew)
```
jgandres:
```suggestion
    if (mState == State::kStateNew)
```

src/chatd.cpp
```diff
@@ -876,36 +868,69 @@ void Connection::doConnect()
...
+
+            mUrl.parse(result->getLink());
+            retryPendingConnection(true);
+        });
+    }
+    else if (disconnect)
+    {
+        CHATDS_LOG_WARNING("retryPendingConnection: forced reconnection!");
+
+        setState(kStateDisconnected);
```
jgandres:
```suggestion
        setState(State::kStateDisconnected);
```

src/chatd.cpp
sergiohs84: No need

PR #582, opened 2019-02-04
Prevent app crashes in logger

jgandres:
Add try-catch for `recursive_mutex::lock()` in order to capture the exception thrown by
the function. There are crash reports from iOS indicating that the static object `Logger` might be already destroyed when `logString()` is called. Apparently, iOS could be destroying the `Logger` object while the `MegaChatApi` instance is still running and logging messages. Hence, the exception upon trying to `lock()` an invalid instance of the mutex.
Due to the high frequency of calls to this function and the overhead introduced by the exception's handler, we will first check another workaround: iOS to force the destroy of the `MegaChatApi` upon app's termination. If the crash still reproduce, then we will merge this PR.

### DIFF
--- a/src/base/logger.cpp
+++ b/src/base/logger.cpp
@@ -2,6 +2,7 @@
 	#define LOGGER_SPRINTF_BUF_SIZE 10240
 #endif
 
+#include <iostream>
 #include <stdarg.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


sergiohs84:
The iOS workaround of destroying `MegaChatApi` upon app's termination haven't fixed the crash :(

PR #602, opened 2019-02-19
Weak references in iOS bindings

jnavarrom:
This way when the app layer set his `MEGAChatSdk` object to `nil`, `MegaChatApiImpl` dtor will be called.


sergiohs84:
Wait for release of groupcalls before merging this branch.

PR #606, opened 2019-02-21
Avoid a full login during SetUp stage

sergiohs84:
In the SetUp stage, MEGAchat's initialization is not needed at all. Better to avoid it, so accounts don't burn that fast :)


sergiohs84:
trigger compilation

PR #612, opened 2019-02-27
MEGAChat initialization statistics

jgandres:


src/chatClient.cpp
```diff
@@ -3663,4 +3731,352 @@ std::string encodeFirstName(const std::string& first)
...
+    mNumNodes = 0;
+    mNumChats = 0;
+    mNumContacts = 0;
+    mTotalElapsed = 0;
+    mFinished = false;
+    mInitState = kInitNewSession;
+    KR_LOG_INFO("Init stats created");
+}
+
+void InitStats::clearStats()
```
sergiohs84: Maybe it's better to declare a `init()`/`reset()` method and call it from ctor and from `onCompleted()`

src/chatClient.cpp
```diff
@@ -3663,4 +3731,352 @@ std::string encodeFirstName(const std::string& first)
...
+{
+    mNumChats = numChats;
+}
+
+bool InitStats::isFinished() const
+{
+    return mFinished;
+}
+
+void InitStats::setFinished(bool finished)
```
sergiohs84: Refactor to `onCompleted()`, without arguments, and add an `assert(!mFinished)`

src/chatClient.cpp
```diff
@@ -1025,6 +1064,7 @@ promise::Promise<void> Client::connect(Presence pres, bool isInBackground)
     }
 
     assert(mConnState == kDisconnected);
+    mInitStats.stageEnd(InitStats::kStatsPostFetchNodes);
```
sergiohs84: This line should go immediately before the `mInitStats.stageStart(InitStats::kStatsConnection);`

PR #613, opened 2019-02-28
Add method to return if client is participating in a call

alber2510:


alber2510:
We've thought that it is better to create other function to return the calls available in one state or other

PR #616, opened 2019-03-05
Set groupcalls always enabled

sergiohs84:


src/chatd.cpp
```diff
@@ -1892,7 +1874,7 @@ void Connection::execCommand(const StaticBuffer& buf)
                 if (mChatdClient.mRtcHandler)
                 {
                     auto& chat = mChatdClient.chats(chatid);
-                    if (!chat.isGroup() || (chat.isGroup() && mChatdClient.mKarereClient->areGroupCallsEnabled()))
+                    if (!chat.isGroup())
```
alber2510: I think we have to remove this condition too. We have to manage CALLTIME for 1to1 and group chatrooms

sergiohs84: You are correct, good catch. Please, can you change it, @alber2510 ?

alber2510: ok

PR #618, opened 2019-03-06
Add functionality for calls

alber2510:
- Modify function to return a all call in the system in a state
-  Return the number of participants with video or audio
- Improve AV flags notifications


alber2510:
trigger compilation

PR #619, opened 2019-03-06
Ensure both users are contacts

sergiohs84:
A 1on1 still exists for ex-contacts, but they cannot operate in the room
unless they become contacts again.

### DIFF
--- a/tests/sdk_test/sdk_test.cpp
+++ b/tests/sdk_test/sdk_test.cpp
@@ -1050,7 +1050,7 @@ void MegaChatApiTest::TEST_EditAndDeleteMessages(unsigned int a1, unsigned int a
     char *secondarySession = login(a2);
 
     MegaUser *user = megaApi[a1]->getContact(mAccounts[a2].getEmail().c_str());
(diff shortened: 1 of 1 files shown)


tests/sdk_test/sdk_test.cpp
```diff
@@ -2557,11 +2555,12 @@ void MegaChatApiTest::TEST_ManualCalls(unsigned int a1, unsigned int a2)
     char *primarySession = login(a1);
     char *secondarySession = login(a2);
 
-    MegaChatHandle uh2 = megaChatApi[a1]->getUserHandleByEmail(mAccounts[a2].getEmail().c_str());
-    if (uh2 == MEGACHAT_INVALID_HANDLE)
+    MegaUser *user = megaApi[a1]->getContact(mAccounts[a2].getEmail().c_str());
+    if (!user || (user->getVisibility() != MegaUser::VISIBILITY_VISIBLE))
```
jgandres:
```suggestion
    if (!user || user->getVisibility() != MegaUser::VISIBILITY_VISIBLE)
```

PR #634, opened 2019-03-22
Fix retrieval of usernames in preview mode

sergiohs84:
In preview mode the unified key is used to encrypt messages, so the public Cu25519 key of senders is not needed, only the public Ed25519 to verify the signature.
(firstname/lastname will be requested at Member's ctor and, for ex-participants, app will request them on demand through MegaChatApi::getUserFirstname/Lastname(userid, ph))

### DIFF
--- a/src/chatClient.cpp
+++ b/src/chatClient.cpp
@@ -1909,6 +1909,7 @@ void GroupChatRoom::initWithChatd(bool isPublic, std::shared_ptr<std::string> un
     }
 
     createChatdChat(users, isPublic, unifiedKey, isUnifiedKeyEncrypted, ph);
+    mChadChatInitialized = true;
 }
 
 void GroupChatRoom::connect(const char *url)
@@ -3275,7 +3276,7 @@ GroupChatRoom::Member::Member(GroupChatRoom& aRoom, const uint64_t& user, chatd:
         {
             self->mRoom.makeTitleFromMemberNames();
         }
-    });
+    }, false, mRoom.mChadChatInitialized ? mRoom.chat().getPublicHandle() : karere::Id::inval().val);
 
     if (!mRoom.parent.mKarereClient.anonymousMode())
     {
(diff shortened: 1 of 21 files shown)


src/chatClient.cpp
```diff
@@ -3132,6 +3132,11 @@ void GroupChatRoom::initChatTitle(const std::string &title, int isTitleEncrypted
     });
 }
 
+bool GroupChatRoom::isChatdChatInitialized()
```
sergiohs84: This method would be more appropriate in `ChatRoom`, despite only groupchats need it by now, but could be usefull for 1on1 in a future.

src/megachatapi_impl.cpp
```diff
@@ -1081,8 +1081,10 @@ void MegaChatApiImpl::sendPendingRequests()
         case MegaChatRequest::TYPE_GET_FIRSTNAME:
         {
             MegaChatHandle uh = request->getUserHandle();
+            const char* publicHandle = request->getLink();
+            MegaChatHandle ph = publicHandle ? karere::Id(publicHandle).val : MEGACHAT_INVALID_HANDLE;
```
sergiohs84: The `Id` ctor should receive the length of the handle as a param (6 bytes). It probably won't make a difference, but it protects against uninitialized bytes in the higher part of the handle.

src/megachatapi.h
```diff
@@ -2618,15 +2618,17 @@ class MegaChatApi
      * The associated request type with this request is MegaChatRequest::TYPE_GET_FIRSTNAME
      * Valid data in the MegaChatRequest object received on callbacks:
      * - MegaChatRequest::getUserHandle - Returns the handle of the user
+     * - MegaChatRequest::getLink - Returns the authorization token
```
sergiohs84: Perhaps we should include something like:
```Previewers of chatlinks are not allowed to retrieve user attributes like firstname or lastname, unless they provide a valid authorization token.```

PR #638, opened 2019-03-27
Fix allow end tailing in `chat-links` URL

jgandres:


sergiohs84:
Replaced by https://github.com/meganz/MEGAchat/pull/640

PR #641, opened 2019-03-28
Fix avoid any callbacks about chatrooms since they are destroyed

jgandres:


polmr:
trigger compilation

alber2510:
trigger compilation

PR #657, opened 2019-04-15
#13141. Avoid conflict in 1to1 calls when two clients answer at same time

alber2510:
If two clients answer an incoming call roughly at same time, a session is established between the 2 clients of the same user (apart from the sessions between the 2 clients and the caller). I could result on the wrong streams being received by the other peer. In example, User A sees video from user B (client 1: B1), but user B in client 2 (B2) sees the video from B1.

**Risk area/s:**
- 1on1 calls answered at two clients at the same time (only one client should success answering the call).
- Group calls answered at two clients of the same user, which is allowed.

### DIFF
--- a/src/rtcModule/webrtc.cpp
+++ b/src/rtcModule/webrtc.cpp
@@ -1056,9 +1056,7 @@ void Call::handleMessage(RtMessage& packet)
     switch (packet.type)
     {
         case RTCMD_CALL_TERMINATE:
-            // It's only used in case of two clients from same user answer in 1to1
-            // chatroom and one of both is finished by the caller
-            msgTerminate(packet);
+            // This message can be received from old clients. It can be ignored
             return;
         case RTCMD_SESSION:
             msgSession(packet);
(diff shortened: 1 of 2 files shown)


src/rtcModule/webrtc.cpp
```diff
@@ -2359,9 +2392,24 @@ void Call::onClientLeftCall(Id userid, uint32_t clientid)
         {
             mManager.launchCallRetry(mChat.chatId(), sentAv(), false);
         }
+    }
+    else if (mState >= kStateInProgress)
+    {
+        if (userid == mManager.mKarereClient.myHandle())
```
sergiohs84: Perhaps we need to add `&& !chat.isGroup()` to the `if` above.

src/rtcModule/webrtc.cpp
```diff
@@ -2359,9 +2392,24 @@ void Call::onClientLeftCall(Id userid, uint32_t clientid)
         {
             mManager.launchCallRetry(mChat.chatId(), sentAv(), false);
         }
+    }
```
sergiohs84: Maybe in the block above we should preserve the `cancelSessionRetryTimer()` and the `destroyIfNoSessionsOrRetries()`?? Not sure, but worths to check it.

src/rtcModule/webrtc.cpp
alber2510: That part of code was added by merge. It's not necessary

src/rtcModule/webrtc.cpp
alber2510: That part of code was added by merge. It's not necessary

sergiohs84:
trigger compilation

sergiohs84:
trigger compilation

alber2510:
trigger compilation

PR #666, opened 2019-05-02
Avoid unhandled promise error

jgandres:


sergiohs84:
trigger compilation

src/chatd.cpp
```diff
@@ -706,6 +706,10 @@ void Connection::setState(State state)
         if (!mSendPromise.done())
         {
             mSendPromise.reject("Failed to send. Socket was closed");
+            mSendPromise.fail([](const ::promise::Error& err)
```
sergiohs84: There is a `.fail()` handler where the promise is created --> in the place where some data is added to the buffer for the socket. If the promise fails, it should, at least, have that handler registered. I can only think a case where the `.fail()` handler is not registered: when the client has started but have not send any data yet, but disconnected. If that's the case, it would be better to start with a promise already _done_ somehow,

PR #675, opened 2019-05-14
Feature #11983: Reset clientid

jgandres:
When the app is running and another instance is launched i.e (share-extension in iOS),
chatd drops the connection if detects that the clientid already exists.

### DIFF
--- a/src/chatClient.cpp
+++ b/src/chatClient.cpp
@@ -645,8 +645,7 @@ promise::Promise<void> Client::initWithNewSession(const char* sid, const std::st
     mMyEmail = getMyEmailFromSdk();
     db.query("insert or replace into vars(name,value) values('my_email', ?)", mMyEmail);
 
-    mMyIdentity = (static_cast<uint64_t>(rand()) << 32) | ::mega::m_time();
-    db.query("insert or replace into vars(name,value) values('clientid_seed', ?)", mMyIdentity);
+    mMyIdentity = initMyIdentity();
 
     mUserAttrCache.reset(new UserAttrCache(*this));
     api.sdk.addGlobalListener(this);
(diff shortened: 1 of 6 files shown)


src/megachatapi.h
```diff
@@ -2185,6 +2185,19 @@ class MegaChatApi
      */
     int init(const char *sid);
 
+    /**
+     * @brief Reset the clientid
```
```suggestion
     * @brief Reset the Client Id for chatd
```

src/megachatapi.h
```diff
@@ -2185,6 +2185,19 @@ class MegaChatApi
...
+     *
+     * When the app is running and another instance is launched i.e (share-extension in iOS),
+     * chatd drops the connection if detects that the clientid already exists.
```
```suggestion
     * chatd closes the connection if a new connection is established with the same client id.
```

src/chatClient.cpp
```diff
@@ -1281,6 +1281,13 @@ uint64_t Client::getMyIdentityFromDb()
     return result;
 }
 
+void Client::resetClientid()
+{
+    KR_LOG_WARNING("Reset the clientid_seed");
+    uint64_t result = (static_cast<uint64_t>(rand()) << 32) | ::mega::m_time();
+    db.query("insert or replace into vars(name,value) values('clientid_seed', ?)", result);
```
sergiohs84: By itself, this method won't update the value in memory for `mMyIdentity`. It needs to be updated.

src/megachatapi_impl.cpp
```diff
@@ -1806,6 +1806,16 @@ int MegaChatApiImpl::init(const char *sid)
     return MegaChatApiImpl::convertInitState(state);
 }
 
+void MegaChatApiImpl::resetClientid()
+{
+    sdkMutex.lock();
+    if (mClient)
+    {
+        mClient->resetClientid();
```
sergiohs84: Perhaps an `assert()` to ensure the initialization state is still waiting-new-session or has-offline-session, but not higher. Otherwise, it means the developer is calling this function too late.

PR #677, opened 2019-05-15
Bug #11953 create new change type to notify apps when a call is fully operative

jgandres:
In mobile apps the "Connecting" label is removed too early, when the
call isn't fully operative so we need to notify apps when the call
is usable.

### DIFF
--- a/src/megachatapi.h
+++ b/src/megachatapi.h
@@ -210,7 +210,7 @@ public:
         CHANGE_TYPE_CALL_COMPOSITION = 0x40,        /// Call composition has changed (User added or removed from call)
         CHANGE_TYPE_SESSION_NETWORK_QUALITY = 0x80, /// Session network quality has changed
         CHANGE_TYPE_SESSION_AUDIO_LEVEL = 0x100,    /// Session audio level has changed
-        CHANGE_TYPE_SESSION_OPERATIVE = 0x200,      /// Session is fully operative
+        CHANGE_TYPE_SESSION_OPERATIVE = 0x200      /// Session is fully operative
     };
 
     enum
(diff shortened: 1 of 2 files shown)


src/megachatapi.h
```diff
@@ -209,7 +209,8 @@ class MegaChatCall
         CHANGE_TYPE_SESSION_STATUS = 0x20,          /// Session status has changed
         CHANGE_TYPE_CALL_COMPOSITION = 0x40,        /// Call composition has changed (User added or removed from call)
         CHANGE_TYPE_SESSION_NETWORK_QUALITY = 0x80, /// Session network quality has changed
-        CHANGE_TYPE_SESSION_AUDIO_LEVEL = 0x100     /// Session audio level has changed
+        CHANGE_TYPE_SESSION_AUDIO_LEVEL = 0x100,    /// Session audio level has changed
+        CHANGE_TYPE_SESSION_OPERATIVE = 0x200,      /// Session is fully operative
```
jgandres:
```suggestion
        CHANGE_TYPE_SESSION_OPERATIVE = 0x200      /// Session is fully operative
```

jgandres:
The callback Session::onAddStream that sends the event FIRE_EVENT(SESS, onVideoRecv) is received for audio-calls too.

src/megachatapi_impl.cpp
```diff
@@ -8010,7 +8010,12 @@ void MegaChatSessionHandler::onPeerMute(karere::AvFlags av, karere::AvFlags oldA
 
 void MegaChatSessionHandler::onVideoRecv()
 {
+    MegaChatCallPrivate *chatCall = callHandler->getMegaChatCall();
+    chatCall->sessionUpdated(session->peer(), session->peerClient(), MegaChatCall::CHANGE_TYPE_SESSION_OPERATIVE);
+    API_LOG_INFO("The call is fully operative. ChatId: %s, callid: %s",
+                 ID_CSTR(chatCall->getChatid()), ID_CSTR(chatCall->getId()));
```
sergiohs84: It would be good to log the `userid` and the `clientid`, since in groupcalls there will be more than one session, and they will receive more than one of this callbacks.

jgandres: I think that It could be useful to notify to each new session to know that a session i s fully operative.

PR #680, opened 2019-05-20
Business accounts maintenance

jgandres:
Notify apps when the status of a business account has changed


sergiohs84:
This PR is rejected. The new event is available to any app using the SDK, no need to wrap it under MEGAchat and duplicate the event's callback.

PR #682, opened 2019-05-24
Adjust types in init stats json

jgandres:
Some variables like total elapsed time are UINT_32 and when json was
generated the variables was represented as a INT_64, wich caused that
the received value was out of range and not corresponds with the
summatory of the rest of the stages.


jgandres:
It's not needed to adjust the data types.

PR #685, opened 2019-05-31
Hide presence status for non-contacts

sergiohs84:
- Hide non-contacts presence status.
- Avoid to send `USERACTIVE 1` if connected in background.
- Add presence config to SettingsWindow in QtApp.

### DIFF
--- a/examples/qtmegachatapi/SettingWindow.cpp
+++ b/examples/qtmegachatapi/SettingWindow.cpp
@@ -28,6 +28,7 @@ SettingWindow::SettingWindow(MegaChatApplication *app) :
 
 SettingWindow::~SettingWindow()
 {
+    delete mPresenceConfig;
     delete ui;
 }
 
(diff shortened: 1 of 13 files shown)


src/presenced.cpp
```diff
@@ -1323,6 +1399,13 @@ void Client::addPeer(karere::Id peer)
 
     if (isExContact(peer))
     {
+        // Notify presence of non-contact that becomes contact again
+        Presence presence = mPeersPresence[peer];
```
alber2510: Check if there's a value for the `peer`. Otherwise, do not insert it in the array. We should use:
```Presence Client::peerPresence(karere::Id peer) const```

Comment (alber2510) [inline src/presenced.cpp]:
```suggestion
        Presence presence = peerPresence(peer);
```

examples/qtmegachatapi/SettingWindow.cpp
```diff
@@ -110,6 +119,27 @@ void SettingWindow::onPushNotificationSettingsUpdate()
     ui->pushNotifications->setEnabled(true);
 }
 
+void SettingWindow::onPresenceConfigUpdate()
+{
+    ui->presence->setEnabled(false);
+
+    delete mPresenceConfig;
```
alber2510: I think mPresenceConfig has to be deleted at SettingWindow destructor too

sergiohs84: You're right. Fixing...

PR #691, opened 2019-06-19
#11541. Streaming renegotiation

alber2510:
It's tested with webRTC commit: 9863f3d246e2da7a2e1f42bbc5757f6af5ec5682

This PR should be merged at the same time than:
- Android's PR: https://github.com/meganz/android2/pull/1048
- iOS PR: https://github.com/meganz/iOS_dev/pull/417
- SDK's PR: https://github.com/meganz/sdk/pull/1755


src/rtcModule/webrtcAdapter.h
```diff
@@ -1,21 +1,23 @@
...
 #include "karereCommon.h" //only for std::string on android
 #include "base/promise.h"
 #include "webrtcAsyncWaiter.h"
 #include "rtcmPrivate.h"
+#include <rtc_base/ref_counter.h>
+
+#ifdef __OBJC__
+@class AVCaptureDevice;
+@class RTCCameraVideoCapturer;
+#else
```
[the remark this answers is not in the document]
javitrujillo: I have just tested, it produces a compilation error:

```
/src/rtcModule/webrtcAdapter.h:404:5: Unknown type name 'AVCaptureDevice'
/src/rtcModule/webrtcAdapter.h:405:5: Unknown type name 'RTCCameraVideoCapturer'
```

src/rtcModule/webrtcAdapter.cpp
```diff
@@ -51,7 +54,18 @@ bool init(void *appCtx)
     threadMgr->SetCurrentThread(thread);
 
     rtc::InitializeSSL();
-    gWebrtcContext = webrtc::CreatePeerConnectionFactory();
+    if (gWebrtcContext == nullptr)
+    {
+        gWebrtcContext = webrtc::CreatePeerConnectionFactory(
```
sergiohs84: We should review the thread's management here, to ensure that it will not break the existing synchronization between MEGAchat's thread and WebRTC's thread/s.

src/rtcModule/webrtcAdapter.h
```diff
@@ -250,190 +255,161 @@ class myPeerConnection: public
...
+    rtc::VideoBroadcaster mBroadcaster;
+    webrtc::MediaSourceInterface::SourceState mState;
+    bool mRemote;
+    rtc::scoped_refptr<webrtc::VideoCaptureModule> mCameraCapturer;
+    webrtc::VideoCaptureCapability mCapabilities;
+    mutable webrtc::webrtc_impl::RefCounter mRefCount{0};
 };
 
-class DeviceManager
+class CapturerTrackSource : public VideoManager
```
sergiohs84: Probably this class can be removed in pro of a design where the `VideoManager` itself inherits from `VideoTrackSourceInterface`, which would result in a cleaner and simpler design.
Furthermore, it should help, as well, to avoid the issue with the destruction of the implementation of such interface, which is an object shared between Objective C and C++ / WebRTC.

src/rtcModule/webrtcAdapter.h
```diff
@@ -250,190 +255,161 @@ class myPeerConnection: public
...
 {
 public:
-    struct InputDevices
-    {
-        DeviceList audio;
-        DeviceList video;
-    };
+    static CapturerTrackSource* Create(const webrtc::VideoCaptureCapability &capabilities, const std::string &deviceName);
+    ~CapturerTrackSource() override;
+    static std::set<std::pair<std::string, std::string>> getVideoDevices();
```
sergiohs84: This method can also fit into the `VideoManager` as an static method.

src/rtcModule/OBJCCaptureModule.mm
```diff
@@ -0,0 +1,157 @@
+
```
sergiohs84: We may want to add this file to the .pro, so it appears in the project inside QtCreator (despite it will not be compiled). It may help to browse the code and to read it, even when its compilation requires xCode

src/rtcModule/webrtcAdapter.h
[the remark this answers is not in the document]
alber2510: Ok, we keep that lines

src/rtcModule/webrtcAdapter.cpp
```diff
@@ -78,171 +88,382 @@ unsigned long generateId()
...
 
-    if (devices.size() == 0)
+    uint32_t numDevices = info->NumberOfDevices();
+    for (uint32_t i = 0; i < numDevices; i++)
     {
-        RTCM_LOG_WARNING("Not suitable input video devices were found. Default device added.");
-        devices.push_back(cricket::Device("default", "0"));
+        char deviceName[256];
+        char uniqueName[256];
+        std::unique_ptr<webrtc::VideoCaptureModule::DeviceInfo> deviceInfo(webrtc::VideoCaptureFactory::CreateDeviceInfo());
```
sergiohs84: Cannot we use the `info` retrieved previously and simply iterate through the items and call `info->GetDeviceName()`??

src/rtcModule/webrtcAdapter.cpp
```diff
@@ -78,171 +88,382 @@ unsigned long generateId()
...
-    MEGAjvm->AttachCurrentThread(&env, NULL);
-    rtc::Thread *currentThread = rtc::ThreadManager::Instance()->CurrentThread();
-    mSource = new rtc::RefCountedObject<webrtc::AndroidVideoTrackSource>(currentThread, env, surfaceTextureHelper, false);
-    rtc::scoped_refptr<webrtc::VideoTrackSourceProxy> proxySource = webrtc::VideoTrackSourceProxy::Create(currentThread, currentThread, mSource);
-    if (startVideoCaptureWithParametersMID)
+    mCameraCapturer->RegisterCaptureDataCallback(this);
+
+    std::unique_ptr<webrtc::VideoCaptureModule::DeviceInfo> deviceInfo(webrtc::VideoCaptureFactory::CreateDeviceInfo());
+    webrtc::VideoCaptureCapability capabilities;
+    deviceInfo->GetCapability(mCameraCapturer->CurrentDeviceName(), 0, capabilities);
```
sergiohs84: Why don't we simply pass here `mCapabilities`?
```suggestion
    deviceInfo->GetCapability(mCameraCapturer->CurrentDeviceName(), 0, mCapabilities);
```
That way we only have to set the `videoType` and we would honor all the capabilities provided by `GetCapability()`, rather than request to `StartCapture()` with our hardcoded capabilities.

src/rtcModule/webrtcAdapter.cpp
```diff
@@ -78,171 +88,382 @@ unsigned long generateId()
...
     {
-        std::string widthString;
-        mOptions->constraints.GetMandatory().FindFirst(webrtc::MediaConstraintsInterface::kMinWidth, &widthString);
-        int width = std::atoi(widthString.c_str());
-        std::string heightString;
-        mOptions->constraints.GetMandatory().FindFirst(webrtc::MediaConstraintsInterface::kMinHeight, &heightString);
-        int height = std::atoi(heightString.c_str());
-        int fps = 15;
-        env->CallStaticVoidMethod(applicationClass, startVideoCaptureWithParametersMID, (jint)width, (jint)height, (jint)fps, (jlong)proxySource.release(), surfaceTextureHelper);
+        return;
```
sergiohs84: We should add an error message to the log here

src/rtcModule/webrtc.cpp
```diff
@@ -2819,16 +2846,26 @@ void Session::veryfySdpOfferSendAnswer()
...
             return;
 
+        uint8_t opcode;
+        if (mState < kStateInProgress)
+        {
+            mTsSdpHandshakeCompleted = time(nullptr);
+            setState(kStateConnecting);
+            opcode = RTCMD_SDP_ANSWER;
+        }
+        else
```
sergiohs84: Should we handle the `mState >= kStateTerminating` here?? Not sure if it's a valid state at this codepath.

src/rtcModule/webrtc.cpp
```diff
@@ -2955,6 +2998,44 @@ void Session::onSignalingChange(webrtc::PeerConnectionInterface::SignalingState
...
+    }
+
+    if (mRenegotiationInProgress)
+    {
+        SUB_LOG_WARNING("Ignoring multiple calls of onRenegotiationNeeded");
+        return;
+    }
+
+    mRenegotiationInProgress = true;
+    setStreamRenegotiationTimeout();
```
sergiohs84: I would move the `mRenegotiationInProgress = true;` into the method that actually sets the corresponding timer. I'll push a commit for that.

src/rtcModule/webrtc.cpp
```diff
@@ -2975,13 +3056,21 @@ void Session::sendAv(AvFlags av)
...
     assert(mPeerAnonId);
     createRtcConn();
+    return sendOffer();
+}
+
+Promise<void> Session::sendOffer()
+{
     auto wptr = weakHandle();
-    return mRtcConn.createOffer(pcConstraints())
+    bool firstOffer = mState < kStateInProgress;
```
sergiohs84: Suggestion: `bool isRenegotiation = mState == kStateInProgress;`
It seems more readable to me. I'll refactorize it.

src/rtcModule/webrtc.cpp
```diff
@@ -2993,25 +3082,35 @@ Promise<void> Session::sendOffer()
...
 
-        SdpKey encKey;
-        mCall.mManager.crypto().encryptKeyTo(mPeer, mOwnHashKey, encKey);
         SdpKey hash;
         mCall.mManager.crypto().mac(mOwnSdpOffer, mPeerHashKey, hash);
 
-        // SDP_OFFER sid.8 anonId.8 encHashKey.32 fprHash.32 av.1 sdpLen.2 sdpOffer.sdpLen
-        cmd(RTCMD_SDP_OFFER,
+        SdpKey encKey;
+        uint8_t command = 0;
```
sergiohs84: This is the `opcode`, not the command, indeed :P 
I'll refactorize it.

src/rtcModule/webrtcPrivate.h
```diff
@@ -64,36 +98,45 @@ class Session: public ISession
     unsigned int mPreviousStatsSize = 0;
     std::unique_ptr<AudioLevelMonitor> mAudioLevelMonitor;
     TermCode mTermCode = TermCode::kInvalid;
+    bool mPeerSupportRenegotiation = false;
+    bool mRenegotiationInProgress = false;
```
sergiohs84: I wonder if this flag should be a new state of the session/call, rather than a state's flag. Let's discuss about it.

webrtc-build/CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ set(WEBRTC_DEFS_LINUX
     -DLINUX -DWEBRTC_POSIX=1 -DHASH_NAMESPACE=__gnu_cxx -DPOSIX -DDISABLE_DYNAMIC_CAST
     -DSSL_USE_NSS -DHAVE_NSS_SSL_H -DSSL_USE_NSS_RNG -DUSE_NSS=1
     -DDYNAMIC_ANNOTATIONS_ENABLED=1 -DWTF_USE_DYNAMIC_ANNOTATIONS=1
-    -DBLINK_SCALE_FILTERS_AT_RECORD_TIME -D_FILE_OFFSET_BITS=64 -DTOOLKIT_VIEWS=1
+    -DBLINK_SCALE_FILTERS_AT_RECORD_TIME -DUNVALID_FILE_OFFSET_BITS=64 -DTOOLKIT_VIEWS=1
```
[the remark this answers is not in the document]
alber2510: I'll remove and check compilation in all platforms

src/rtcModule/webrtc.cpp
```diff
@@ -2955,6 +2998,49 @@ void Session::onSignalingChange(webrtc::PeerConnectionInterface::SignalingState
...
+    }
+}
+
+void Session::onRenegotiationNeeded()
+{
+    if (mState != kStateInProgress)
+    {
+        if (mState < kStateInProgress)
+        {
+            mRenegotiationAfterInitialConnect = true;
```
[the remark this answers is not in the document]
alber2510: It's ok

src/rtcModule/webrtcAdapter.cpp
sergiohs84: The `mCapabilities` is already fixed to the wanted video resolution and fps. Only the `interlaced` capability needs to be the same than the supported by the device. The video format is preset to `I420` because that's the output video format, so we can avoid the conversion of frames from another format.

src/rtcModule/webrtcAdapter.cpp
alber2510: Fixed

src/rtcModule/webrtc.cpp
```diff
@@ -221,16 +146,49 @@ void RtcModule::removeCallRetry(karere::Id chatid)
...
+    std::set<std::pair<std::string, std::string>> videoDevices = loadDeviceList();
+    for (const std::pair<std::string, std::string> &device : videoDevices)
+    {
+        if (mVideoDeviceSelected == device.second)
+        {
+            return device.first;
+        }
+    }
+
+    return std::string();
```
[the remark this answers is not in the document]
alber2510: Fixed, by protecting _openDevice_

src/rtcModule/webrtc.cpp
```diff
@@ -2223,6 +2100,105 @@ void Call::monitorCallSetupTimeout()
...
+
+void Call::enableVideo(bool enable)
+{
+    if (mState >= Call::kStateTerminating)
+    {
+        return;
+    }
+
+    if (enable)
+    {
```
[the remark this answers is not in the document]
alber2510: It's created at beginning of the call

src/rtcModule/webrtc.cpp
alber2510: It's checked few lines above

src/rtcModule/webrtcPrivate.h
alber2510: It's not really necessary, session is terminating when the flag is set

webrtc-build/CMakeLists.txt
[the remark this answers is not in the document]
alber2510: The flags are not necessary

PR #692, opened 2019-06-21
Cancel statistics when apps goes to background

jgandres:
In order to discard samples with atypical values because of inactivity periods, we need to cancel statistics when the app tries to connect in background or when the apps goes to background.

### DIFF
--- a/src/chatClient.cpp
+++ b/src/chatClient.cpp
@@ -4111,7 +4111,7 @@ std::string InitStats::toJson()
     jSonObject.AddMember(rapidjson::Value("sid"), jsonValue, jSonDocument.GetAllocator());
 
     // Add init stats version
-    uint32_t version = 2;
+    uint32_t version = INITSTATSVERSION;
     jsonValue.SetUint(version);
     jSonObject.AddMember(rapidjson::Value("v"), jsonValue, jSonDocument.GetAllocator());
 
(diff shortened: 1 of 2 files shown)


src/chatClient.cpp
```diff
@@ -3924,6 +3943,7 @@ void InitStats::stageStart(uint8_t stage)
     }
 
     mStageStats[stage] = currentTime();
+    KR_LOG_WARNING("Stage: %d Start: %ld", stage, mStageStats[stage]);
```
alber2510:
```suggestion
    KR_LOG_DEBUG("Stage: %d Start: %ld", stage, mStageStats[stage]);
```

src/chatClient.cpp
```diff
@@ -3934,7 +3954,9 @@ void InitStats::stageEnd(uint8_t stage)
     }
 
     assert(mStageStats[stage]);
-    mStageStats[stage] = currentTime() - mStageStats[stage];
+    mega::dstime aux = currentTime();
+    mStageStats[stage] = aux - mStageStats[stage];
+    KR_LOG_WARNING("Stage: %d End: %ld Elap: %ld", stage, aux, mStageStats[stage]);
```
alber2510:
```suggestion
    KR_LOG_DEBUG("Stage: %d End: %ld Elap: %ld", stage, aux, mStageStats[stage]);
```

PR #695, opened 2019-06-28
Update for SDK moving to std::mutex

mattw-mega:
Also addressed the most verbose of the template warnings.


src/chatClient.h
```diff
@@ -14,10 +14,20 @@
 #include "IGui.h"
 #include <base/trackDelete.h>
 #include "rtcModule/webrtc.h"
+#include "stringUtils.h"
+
+#ifdef _WIN32
```
cb-mega: I'd consolidate this rapidjson stuff into a single header and then include that everywhere

PR #700, opened 2019-07-08
Hotfix #12611. Launch call reconnection at states in-progress and joining

alber2510:


sergiohs84:
Let's wait for Android. If they do not release a patched version based on https://github.com/meganz/MEGAchat/tree/v2.1.4, better to merge this fix only into `develop` through https://github.com/meganz/MEGAchat/pull/699

PR #704, opened 2019-07-22
#12551 Use nickname if available

jgandres:
Wherever a firstname / fullname is currently displayed, if the user has assigned to him/her a nickname, it should prevail. In example, default groupchat title, notifications about new messages, sender of a message, etc.

### DIFF
--- a/src/chatClient.cpp
+++ b/src/chatClient.cpp
@@ -2153,8 +2153,8 @@ void PeerChatRoom::initContact(const uint64_t& peer)
 
             // If the contact has alias don't update the title
             auto self = static_cast<PeerChatRoom*>(userp);
-            const char *alias = self->parent.mKarereClient.getUserAlias(self->mPeer);
-            if (!alias)
+            std::string alias = self->parent.mKarereClient.getUserAlias(self->mPeer);
+            if (alias.size() <= 1)
             {
                 if (!data || data->empty() || (*data->buf() == 0 && data->size() == 1))
                 {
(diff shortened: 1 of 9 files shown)


src/chatClient.cpp
```diff
@@ -3789,6 +3850,88 @@ bool Client::isCallInProgress(karere::Id chatid) const
...
+{
+    std::string aliasBin;
+    std::map<uint64_t, std::string>::iterator it;
+    it = mAliasesMap.find(userId);
+    if (it != mAliasesMap.end())
+    {
+        std::string aliasB64 = it->second;
+        Buffer buf(aliasB64.size());
+        size_t decLen = base64urldecode(aliasB64.c_str(), aliasB64.size(), buf.buf(), buf.bufSize());
+        aliasBin.assign(buf.buf(), decLen);
```
sergiohs84: All the above for a simple conversion from B64 to binary seems too complicated for a simple operation. Please, pay attention to my adjustment.

src/userAttrCache.cpp
```diff
@@ -439,8 +428,14 @@ void UserAttrCache::fetchStandardAttr(UserAttrPair key, std::shared_ptr<UserAttr
...
-        item->data.reset(gUserAttrDescs[key.attrType].getData(*result));
-        item->resolve(key);
+
+        UserAttrDescMap::iterator it = gUserAttrDescsMap.find(key.attrType);
+        if (it != gUserAttrDescsMap.end())
+        {
+            auto& desc = it->second;
+            item->data.reset(desc.getData(*result));
+            item->resolve(key);
+        }
```
sergiohs84: Is it possible to fetch an attribute that it's not at the `gUserAttrDescsMap`?? I think it shouldn't. Let's discuss about it.

sergiohs84:
trigger compilation

src/userAttrCache.cpp
jgandres: I agree with you, we shouldn't fetch an attribute out of gUserAttrDescsMap

sergiohs84: Then I'll simplify it :)

sergiohs84:
TODO:
- Implement the deletion of aliases in `updateAliases()`.
- Add the _fullname_ to the contact in the contact list of the QtApp (if there's an alias).

PR #705, opened 2019-08-07
#12877 Not manage as rich link mega urls

alber2510:
Some services as facebook introduce a special string in mega urls. Detect this special cases and avoid to manage as rich link

### DIFF
--- a/src/chatd.cpp
+++ b/src/chatd.cpp
@@ -5094,7 +5094,7 @@ bool Message::parseUrl(const std::string &url)
         }
     }
 
-    std::regex megaUrlExpression("^((WWW.|www.)?mega.+(nz/|co.nz/)).*((#F!|#!|chat/)[a-z0-9A-Z-._~:\/?#!$&'()*+,;=\-@]+)$");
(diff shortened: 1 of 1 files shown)


alber2510:
trigger compilation

src/chatd.cpp
```diff
@@ -5094,16 +5094,14 @@ bool Message::parseUrl(const std::string &url)
         }
     }
 
-    if (urlToParse.find("mega.co.nz/#!") != std::string::npos || urlToParse.find("mega.co.nz/#F!") != std::string::npos ||
-            urlToParse.find("mega.nz/#!") != std::string::npos || urlToParse.find("mega.nz/#F!") != std::string::npos ||
-            urlToParse.find("mega.nz/chat/") != std::string::npos)
+    std::regex megaUrlExpression("^((WWW.|www.)?mega.+(nz/|co.nz/)).*((#F!|#!|chat/)[a-z0-9A-Z-._~:\/?#!$&'()*+,;=\-@]+)$");
```
sergiohs84: Perhaps we should skip also the contact links: `https://mega.nz/C!<public_handle>`

sergiohs84: Apps will eventually show an enhanced UI for contact-links. Hence, the library should stop creating rich-links for that kind of links.
I'll fix it

PR #706, opened 2019-08-09
#12043. Chat reactions

jgandres:


src/strongvelope/strongvelope.cpp
```diff
@@ -624,6 +624,30 @@ ProtocolHandler::ProtocolHandler(karere::Id ownHandle,
     }
 }
 
+template<typename T>
+std::vector<T> ProtocolHandler::str_to_a32(std::string data)
```
sergiohs84: Check if this method provides the same result than `DataTransformation::vector_to_b()` from `megachatapi_impl.cpp` (you can test by sending & receiving a node-attachment and downloading it at both sides). If they are compatible, remove the former one in pro of the latter one.

Also check if it is compatible with the analogous in the SDK, `MegaClient::str_to_a32()` (check the password hash generation in Webclient's code and/or set a new password including some UTF8 chars). If they are compatible, replace the function/s in the SDK with these ones and use them everywhere.

src/chatd.cpp
```diff
@@ -1705,6 +1755,8 @@ Chat::Chat(Connection& conn, Id chatid, Listener* listener,
     mLastReceivedId = info.lastRecvId;
     mLastSeenIdx = mDbInterface->getIdxOfMsgidFromHistory(mLastSeenId);
     mLastReceivedIdx = mDbInterface->getIdxOfMsgidFromHistory(mLastReceivedId);
+    std::string rsn = (!previewMode()) ? mDbInterface->getReactionSn() : std::string();
```
sergiohs84: No need to check the preview mode, but to save the chat in cache at the GroupChatRoom::ctor earlier than this method is called.
I'll fix it.

src/chatd.cpp
```diff
@@ -4743,6 +5001,9 @@ void Chat::setOnlineState(ChatState state)
 
     if (state == kChatStateOnline)
     {
+        // Send REACTIONSN after a reconnection
+        sendReactionSn();
```
sergiohs84: It should be sent even before the `HIST`/`JOINRANGEHIST`/`HANDLEJOIN`...
I'll fix it

examples/qtmegachatapi/reaction.cpp
```diff
@@ -0,0 +1,144 @@
...
+}
+
+std::string Reaction::getReactionString() const
+{
+    return mReactionString;
+}
+
+void Reaction::updateReactionCount(int count)
+{
+    if (!count)
```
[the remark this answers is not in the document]
jgandres: In case reaction count is 0 the widget will be removed in ChatMessage::updateReaction, so we don't need to update the text.

src/chatd.cpp
```diff
@@ -4688,6 +4833,104 @@ void Chat::onUserLeave(Id userid)
...
+        mCrypto->reactionDecrypt(message, reaction)
+        .then([this, wptr, message, userId](std::shared_ptr<Buffer> data)
+        {
+            if (wptr.deleted())
+                return;
+
+            std::string reaction (data->buf(), data->bufSize());
+            message->addReaction(reaction, userId);
+
+            if (!previewMode())
```
sergiohs84: Can a chat in preview mode upgrade due to auto-join? In that case, does the chatroom created during the preview preserved, or is it destroyed and a new is created from scratch? If the answer if the former, then you do need to add reactions to cache.
I'll fix it

sergiohs84: Perhaps it's a good idea to move the addition to DB in the `Message::addReaction()` (same for `delReaction()`). TBD based on the logic of the pending-reactions queue

examples/qtmegachatapi/reaction.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+void Reaction::onCopyReact()
+{
+    QClipboard *clipboard = QApplication::clipboard();
+    clipboard->setText(mReactionString.c_str());
+}
+
+void Reaction::onRemoveReact()
+{
+    ChatWindow *chatwindow = mChatMessage->getChatWindow();
```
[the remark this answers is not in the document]
sergiohs84: If you prefer to register the slot from ChatMessage directly for the signal captured in Reaction, I support it :)

src/megachatapi.h
```diff
@@ -5469,6 +5568,7 @@ class MegaChatRoomListener
      * @param chat MegaChatRoom whose local history is about to be discarded
      */
     virtual void onHistoryReloaded(MegaChatApi* api, MegaChatRoom *chat);
+    virtual void onReactionUpdate(MegaChatApi* /*api*/, MegaChatHandle /*msgid*/, const char* /*reaction*/, int /*count*/);
```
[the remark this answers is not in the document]
sergiohs84: I've fixed it already ;)

PR #712, opened 2019-08-30
#13115. Fix isManagementMessage

jgandres:
For public chats keyid is 0 for all messages, so we can't check keyid
to identify if a message is a management message or not.

To solve this problem now we will check if the userid is the API user,
except for truncate messages and if message type is a management known
type.

### DIFF
--- a/src/chatdMsg.h
+++ b/src/chatdMsg.h
@@ -675,6 +675,7 @@ public:
     /** @brief Returns whether this message is a management message. */
     bool isManagementMessage() const
     {
+        // if message comes from API and uses keyid=0, it's a management message
         return isManagementMessageKnownType() || (userid == karere::Id::COMMANDER() && keyid == 0);
     }
     bool isManagementMessageKnownType() const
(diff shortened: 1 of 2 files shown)


src/chatdMsg.h
```diff
@@ -675,12 +675,12 @@ class Message: public Buffer
     /** @brief Returns whether this message is a management message. */
     bool isManagementMessage() const
     {
-        return (keyid == 0) && !isSending();    // msgs in sending status use keyid=CHATD_KEYID_INVALID (0)
+        return (userid == karere::Id::COMMANDER() || type == Message::kMsgTruncate)
```
sergiohs84: The `userid` is overwritten when it goes through `strongvelope::msgDecrypt()`. It is set to the user who originates the management message, so here the condition will only be satisfied for truncates.

sergiohs84: The right condition could be:
```
return isManagementMessageKnownType() || (userid == karere::Id::COMMANDER() && keyid == 0);
```

PR #714, opened 2019-09-11
megacli: commands to test get/set camera folder

mattw-mega:


examples/megaclc/megaclc.cpp
```diff
@@ -3239,6 +3240,45 @@ void exec_getmegaachievements(ac::ACState& s)
     g_megaApi->getMegaAchievements(listener);
 }
 
+void exec_setCameraUploadsFolder(ac::ACState& s)
+{
+    std::unique_ptr<m::MegaNode> srcnode(g_megaApi->getNodeByPath(s.words[1].s.c_str()));
+    
+    if (!srcnode)
+    {
+        cout << "folder not found";
```
sergiohs84: Should we use a `conlock(cout)` here?
```suggestion
        conlock(cout) << "Folder not found: " << s.words[1].s << endl;
```

examples/megaclc/megaclc.cpp
```diff
@@ -3239,6 +3240,45 @@ void exec_getmegaachievements(ac::ACState& s)
...
+void exec_getCameraUploadsFolder(ac::ACState& s)
+{
+    g_megaApi->getCameraUploadsFolder(new OneShotRequestListener([](m::MegaApi*, m::MegaRequest *r, m::MegaError* e)
+    {
+        if (check_err("getCameraUploadsFolder", e, ReportFailure))
+        {
+            unique_ptr<m::MegaNode> node(g_megaApi->getNodeByHandle(r->getNodeHandle()));
+            if (!node)
+            {
+                conlock(cout) << "No node found by looking up handle: '" << r->getNodeHandle() << "'" << endl;
```
sergiohs84: Perhaps it's more useful to print handles in `B64`.

examples/megaclc/megaclc.cpp
mattw-mega: good point, fixed :-)

examples/megaclc/megaclc.cpp
mattw-mega: Yes, quite right.  I've added a function to make that convenient, and made the fix. thanks!

PR #719, opened 2019-09-23
#13283. Fix QtApp construction by adding libraw as dependency

jgandres:
The SDK doesn't need libraw to work, but it improves UX because it generates previews of RAW format images. Altough is not specified in configure, HAVE_LIBRAW is added to config.h as long as libraw is available in the system.

We need to add libraw as QtApp dependency.

**Risk area/s:**
- QtApp construction


polmr:
trigger compilation

PR #721, opened 2019-09-25
#13040. Retry pending reactions

jgandres:
Ready to review


jgandres:
trigger compilation

jgandres:
trigger compilation

src/chatd.h
```diff
@@ -1602,12 +1643,17 @@ class DbInterface
     virtual Idx getIdxOfMsgidFromNodeHistory(karere::Id msgid) = 0;
 
     //  <<<--- Reaction methods --->>>
-    virtual std::string getReactionSn() = 0;
+    virtual const std::string getReactionSn() const = 0;
     virtual void setReactionSn(const std::string &rsn) = 0;
     virtual void cleanReactions(karere::Id msgId) = 0;
+    virtual void cleanPendingReactions(karere::Id msgId) = 0;
     virtual void addReaction(karere::Id msgId, karere::Id userId, const char *reaction) = 0;
+    virtual void addPendingReaction(karere::Id msgId, const char *reaction, std::string encReaction, uint8_t status) = 0;
```
alber2510, line 1651: I think it's always better use `const std::string&` than `const char* ` as parameter.
My proposal would be
`virtual void addPendingReaction(karere::Id msgId, const std::string& reaction, const std::string& encReaction, uint8_t status) = 0;`

But it's only my opinion

src/chatd.cpp
```diff
@@ -5457,6 +5591,14 @@ Chat::ManualSendItem::ManualSendItem()
 
 }
 
+Chat::PendingReaction::PendingReaction(std::string aReactionString, std::string aReactionStringEnc, uint64_t aMsgId, uint8_t aStatus)
+{
+    mMsgId = aMsgId;
+    mStatus = aStatus;
+    mReactionStringEnc = std::move(aReactionStringEnc);
```
alber2510, line 5598: I think if we use `const referece` as parameter `std::move` aren't necessary

src/chatd.cpp
```diff
@@ -2581,28 +2596,133 @@ void Chat::sendSync()
...
-           std::string encReaction (data->buf(), data->bufSize());  // lenght must be only 1 byte. passing the buffer uses 4 bytes for size
-           sendCommand(Command(OP_ADDREACTION) + mChatId + client().myHandle() + message.id() + (int8_t)data->bufSize() + encReaction);
-        })
-        .fail([this](const ::promise::Error& err)
+void Chat::retryPendingReactions()
+{
+    for (auto& reaction: mPendingReactions)
+    {
+        karere::Id msgid = reaction.mMsgId;
+        Idx index = msgIndexFromId(msgid);
```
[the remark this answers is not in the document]
jgandres: You shouldn't be able to react to messages that are not displayed in GUI so these messages must be loaded in RAM.

jgandres:
trigger compilation

src/chatd.cpp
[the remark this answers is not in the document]
alber2510: Yes, you are right :+1:

src/megachatapi_impl.cpp
```diff
@@ -4467,70 +4485,121 @@ MegaChatErrorPrivate *MegaChatApiImpl::delReaction(MegaChatHandle chatid, MegaCh
...
 MegaStringList* MegaChatApiImpl::getMessageReactions(MegaChatHandle chatid, MegaChatHandle msgid)
-{    
+{
     SdkMutexGuard g(sdkMutex);
-
-    vector<char*> reactArray;
+    ChatRoom *chatroom = findChatRoom(chatid);
     Message *msg = findMessage(chatid, msgid);
-    if (msg)
+    if (!msg || !chatroom)
```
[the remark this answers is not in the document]
sergiohs84: I've revamped the code accordingly here: https://github.com/meganz/MEGAchat/pull/721/commits/56fc854550d73e1facbd8f29f8b0482533d5c26c

src/chatdDb.h
```diff
@@ -368,6 +376,9 @@ class ChatdSqliteDb: public chatd::DbInterface
         // Clean reactions for the truncate message
         mDb.query("delete from chat_reactions where chatid = ? and msgid = ?", mChat.chatId(), msg.id());
 
+        // Clean pending reactions for the truncate message
+        mDb.query("delete from chat_pending_reactions where chatid = ? and msgid = ?", mChat.chatId(), msg.id());
```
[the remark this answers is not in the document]
sergiohs84: Done!

src/megachatapi_impl.cpp
```diff
@@ -4467,70 +4485,121 @@ MegaChatErrorPrivate *MegaChatApiImpl::delReaction(MegaChatHandle chatid, MegaCh
...
     {
-        std::vector<std::string> reactions(msg->getReactions());
-        for (auto &it : reactions)
+        API_LOG_ERROR("Chatroom or message not found");
+        return new MegaStringListPrivate();
+    }
+
+    // Insert confirmed reactions
+    std::map<std::string, int> auxReactMap;
+    const std::vector<std::string> &reactions = msg->getReactions();
```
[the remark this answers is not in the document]

src/chatd.cpp
```diff
@@ -2667,28 +2684,121 @@ void Chat::sendSync()
...
+    for (auto& reaction: mPendingReactions)
+    {
+        assert(!reaction.mReactionStringEnc.empty());
+        sendCommand(Command(reaction.mStatus) + mChatId + client().myHandle() + reaction.mMsgId + (int8_t)reaction.mReactionStringEnc.size() + reaction.mReactionStringEnc);
+    }
+}
+
+void Chat::cleanPendingReactionsByMsg(karere::Id msgid, Idx idx)
+{
+    for (auto it = mPendingReactions.begin(); it != mPendingReactions.end();)
```
sergiohs84, line 2745: There is a very handy method for `std::list` --> `remove_if`
http://www.cplusplus.com/reference/list/list/remove_if/

We should take advantage of those algorithms when possible, since they are easier to read and usually more efficient. The implementation, in example, for a similar method:
``` cpp
void Chat::cleanPendingReactions(const karere::Id &msgId)
{
    mPendingReactions.remove_if([msgId](PendingReaction& reaction)
    {
        return reaction.mMsgId == msgId;
    });
}
```

src/chatd.cpp
```diff
@@ -2667,28 +2684,121 @@ void Chat::sendSync()
...
+            // couldn't find msg
+            CHATID_LOG_WARNING("flushChatPendingReactions: message with id(%d) not loaded in RAM", msgid);
+        }
+        else
+        {
+            Message &message = at(index);
+            CALL_LISTENER(onReactionUpdate, message.mId, auxit->mReactionString.c_str(), message.getReactionCount(auxit->mReactionString.c_str()));
+        }
+        CALL_DB(cleanPendingReactions, auxit->mMsgId);
+        mPendingReactions.erase(auxit);
```
sergiohs84: Instead of erasing item by item, you can iterate as usual, with a `for (auto &reaction : mPendingReactions)` and after the loop, do a `mPendingReactions.clear()`

src/chatd.cpp
```diff
@@ -2667,28 +2684,121 @@ void Chat::sendSync()
...
+        Idx index = msgIndexFromId(msgid);
+        if (index == CHATD_IDX_INVALID)
+        {
+            // couldn't find msg
+            CHATID_LOG_WARNING("flushChatPendingReactions: message with id(%d) not loaded in RAM", msgid);
+        }
+        else
+        {
+            Message &message = at(index);
+            CALL_LISTENER(onReactionUpdate, message.mId, auxit->mReactionString.c_str(), message.getReactionCount(auxit->mReactionString.c_str()));
```
sergiohs84: You are passing a `c_str()` to `getReactionCount()`, when it accepts a const ref to string. Suboptimal. I'll fix it.

src/chatd.cpp
```diff
@@ -5229,8 +5340,7 @@ void Chat::onUserLeave(Id userid)
 
 void Chat::onAddReaction(Id msgId, Id userId, std::string reaction)
 {
-    Idx messageIdx = msgIndexFromId(msgId);
-    if (messageIdx == CHATD_IDX_INVALID)
+    if (mDbInterface->getIdxOfMsgidFromHistory(msgId) == CHATD_IDX_INVALID)
```
sergiohs84, line 5337: Searching for the `msgId` in the DB should be the last option. It's slower than memory, and most of the times the reaction corresponds to a message in memory.

src/chatd.cpp
```diff
@@ -5270,8 +5410,7 @@ void Chat::onAddReaction(Id msgId, Id userId, std::string reaction)
 
 void Chat::onDelReaction(Id msgId, Id userId, std::string reaction)
 {
-    Idx messageIdx = msgIndexFromId(msgId);
-    if (messageIdx == CHATD_IDX_INVALID)
+    if (mDbInterface->getIdxOfMsgidFromHistory(msgId) == CHATD_IDX_INVALID)
```
sergiohs84, line 5407: Same than for `onAddReaction()`. Not efficient to look up at DB for every reaction just to check if the `msgid` is locally available or not. Please, if really needed, find a more efficient way to check it.

src/chatd.h
```diff
@@ -1695,12 +1738,17 @@ class DbInterface
...
     virtual void cleanReactions(karere::Id msgId) = 0;
-    virtual void addReaction(karere::Id msgId, karere::Id userId, const char *reaction) = 0;
-    virtual void delReaction(karere::Id msgId, karere::Id userId, const char *reaction) = 0;
-    virtual void getMessageReactions(karere::Id msgId, ::mega::multimap<std::string, karere::Id>& reactions) = 0;
+    virtual void cleanPendingReactions(karere::Id msgId) = 0;
+    virtual void addReaction(karere::Id msgId, karere::Id userId, const std::string &reaction) = 0;
+    virtual void addPendingReaction(karere::Id msgId, const std::string &reaction, const std::string &encReaction, uint8_t status) = 0;
+    virtual void delReaction(karere::Id msgId, karere::Id userId, const std::string &reaction) = 0;
+    virtual void delPendingReaction(karere::Id msgId, const std::string &reaction) = 0;
+    virtual void getMessageReactions(karere::Id msgId, std::multimap<std::string, karere::Id>& reactions) const = 0;
```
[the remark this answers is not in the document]
sergiohs84: Done

src/chatd.h
```diff
@@ -1339,7 +1350,34 @@ class Chat: public karere::DeleteTrackable
...
+
+    /** @brief Clean pending reactions in a chat for a specific message or range of messages.
+     *
+     * - If idx is not valid the function will clean pending reactions for the given msgid.
+     * - Otherwise, the function will clean all pending reactions for the messages whose
+     * index is previous to idx.
+     *
+     * Reactions in local DB are removed along with messages (FK delete on cascade).
+     */
+    void cleanPendingReactionsByMsg(karere::Id msgid, Idx idx);
```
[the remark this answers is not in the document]
sergiohs84: Done

src/chatd.cpp
sergiohs84: Done!

src/chatd.cpp
```diff
@@ -5783,6 +5951,14 @@ Chat::ManualSendItem::ManualSendItem()
 
 }
 
+Chat::PendingReaction::PendingReaction(const std::string &aReactionString, const std::string &aReactionStringEnc, uint64_t aMsgId, uint8_t aStatus)
```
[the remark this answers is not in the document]
sergiohs84: Done

src/chatd.cpp
```diff
@@ -1911,11 +1916,12 @@ Idx Chat::getHistoryFromDb(unsigned count)
...
         // Load msg reactions from cache
-        ::mega::multimap<std::string, karere::Id> reactions;
+        std::multimap<std::string, karere::Id> reactions;
         CALL_DB(getMessageReactions, msg->id(), reactions);
-        for (auto &it : reactions)
+        for (auto auxReact : reactions)
         {
-            msg->addReaction(it.first, it.second);
+            // Add reaction to confirmed reactions queue in message
+            msg->addReaction(auxReact.first.c_str(), auxReact.second);
```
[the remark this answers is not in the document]
sergiohs84: Fixed

src/chatd.cpp
sergiohs84: Fixed

src/chatd.cpp
sergiohs84: Fixed

sergiohs84:
Apart from the comments in the PR, @jgandres (please, check the resolved ones too, they are important too):
- The QtApp does not show any reaction after you close the ChatWindow and reopen. It doesn't either when you close the app and relaunch it. Only shows the reactions generated while the ChatWindow is open. Maybe, an issue after the changes for retention-history??
- I wonder what happens, regarding pending reactions, if client receives a `REJECT` for a `JOINRANGEHIST`, which force the client to discard local history and refetch from scratch.
- We should double-check that `JOINRANGENODEHIST` (for loading attachments) do not result on `ADD/DELREACTION`s for received messages. It would be a bug in chatd, it's worthy to check the corner case.

src/chatdDb.h
```diff
@@ -368,6 +376,9 @@ class ChatdSqliteDb: public chatd::DbInterface
         // Clean reactions for the truncate message
         mDb.query("delete from chat_reactions where chatid = ? and msgid = ?", mChat.chatId(), msg.id());
```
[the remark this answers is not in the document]
jgandres: Truncate msg is not removed from db, so we need to clean it's reactions

src/chatd.cpp
jgandres: Solved in e71ebaf112242c882497ef3fd95b2d7d9b99ef49

jgandres:
>     * The QtApp does not show any reaction after you close the ChatWindow and reopen. It doesn't either when you close the app and relaunch it. Only shows the reactions generated while the ChatWindow is open. Maybe, an issue after the changes for retention-history??

Fixed in: 24cae46db20163c9761b817915d9b28620319ba6

src/chatd.cpp
```diff
@@ -2778,36 +2778,20 @@ void Chat::flushChatPendingReactions()
...
-    addPendingReaction(reaction, encReaction, message.id(), OP_DELREACTION);
-    CALL_DB(addPendingReaction, message.mId, reaction, encReaction, OP_DELREACTION);
-    auto wptr = weakHandle();
-    marshallCall([wptr, this, &message, data, encReaction]()
-    {
-        if (wptr.deleted())
-            return;
-
-        sendCommand(Command(OP_DELREACTION) + mChatId + client().myHandle() + message.id()
+        sendCommand(Command(opcode) + mChatId + client().myHandle() + message.id()
```
sergiohs84, line 2794: We still have a potential race-condition with this asynchronous call. You don't have any control on when this command will be sent, not even if it will be sent (due to network being down). Hence, the sequence of opcodes sent to chatd may not correspond to the expected sequence by the user, resulting on the incorrect state of the reaction.

A possible workaround is to convert the `MegaChatApi::add/delReaction()` into asynchronous requests. The required sanity checks would happen in the karere's thread and the command can be sent synchronously in the `sendPendingRequests()` directly. That way, you avoid any race and honor the order of the queue of pending reactions.

jgandres: Solved in PR: #907

PR #736, opened 2019-10-21
#13644 megaclc: added test commands for get/setcamerauploadfoldersecondary.

mattw-mega:
needs the SDK branch of the same name.


mattw-mega:
trigger compilation

sergiohs84:
BTW, I've tested this PR with the new proposal in the SDK and it works fine :)

PR #739, opened 2019-10-31
#13733. Fix history reload upon JOINRANGEHIST REJECT

jgandres:
**Risk area/s:**
- Load history from server

### DIFF
--- a/src/chatd.cpp
+++ b/src/chatd.cpp
@@ -1612,7 +1612,7 @@ HistSource Chat::getNodeHistory(uint32_t count)
     return mAttachmentNodes->getHistory(count);
 }
 
-HistSource Chat::getHistoryFromDbOrServer(unsigned count)
+HistSource Chat::getHistoryFromDbOrServer(unsigned count, bool initialFetch)
 {
     if (mHasMoreHistoryInDb)
     {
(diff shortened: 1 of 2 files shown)


src/chatd.cpp
```diff
@@ -2081,7 +2083,7 @@ void Connection::execCommand(const StaticBuffer& buf)
                     chat.clearHistory();
                     // we were notifying NEWMSGs in result of JOINRANGEHIST, but after reload we start receiving OLDMSGs
                     chat.mServerOldHistCbEnabled = mChatdClient.mKarereClient->isChatRoomOpened(chatid);
-                    chat.getHistoryFromDbOrServer(chat.initialHistoryFetchCount);
+                    chat.getHistoryFromDbOrServer(chat.initialHistoryFetchCount, true);
```
sergiohs84: I think it would be better to solve this bug by:
```suggestion
                    chat.getHistoryFromServer(-chat.initialHistoryFetchCount);
```
No more changes would be required, and the state variables would remain unchanged (like `mServerOldHistCbEnabled`, which must be `false` if the chatroom is not opened at that moment.

jgandres: I agree with you

jgandres:
trigger compilation

PR #753, opened 2019-12-16
#14071. Avoid race conditions on DelegateChatRoomListener

sergiohs84:
Callbacks from MegaChatRoomListener are queued by DelegateChatRoomListener in order to be executed by the app's thread. In result, the execution may happen when the chatroom has been closed.

Since the MegaChatRoomListenerInterface is reused by the app for a new DelegateChatRoomListener when another chatroom is opened, it may receive outdated callbacks, corresponding to the already closed chatroom.

This commit aims to avoid such race condition by properly invalidating the listener when the chatroom is closed.

### DIFF
--- a/bindings/java/nz/mega/sdk/MegaChatApiJava.java
+++ b/bindings/java/nz/mega/sdk/MegaChatApiJava.java
@@ -1750,6 +1750,8 @@ public class MegaChatApiJava {
             DelegateMegaChatRoomListener item = itr.next();
             if(item.getUserListener() == listener){
                 listenerToDelete = item;
+                boolean success = listenerToDelete.invalidateUserListener();
+                assert success; // failed if listener was already invalidated
                 itr.remove();
                 break;
             }
--- a/bindings/java/nz/mega/sdk/DelegateMegaChatRoomListener.java
+++ b/bindings/java/nz/mega/sdk/DelegateMegaChatRoomListener.java
@@ -30,6 +30,14 @@ class DelegateMegaChatRoomListener extends MegaChatRoomListener {
     MegaChatRoomListenerInterface getUserListener() {
         return listener;
     }
+    
+    boolean invalidateUserListener() {
+		if (listener != null) {
+			listener = null;
+			return true;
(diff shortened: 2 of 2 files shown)


bindings/java/nz/mega/sdk/MegaChatApiJava.java
```diff
@@ -1750,6 +1750,8 @@ public void closeChatRoom(long chatid, MegaChatRoomListenerInterface listener){
             DelegateMegaChatRoomListener item = itr.next();
             if(item.getUserListener() == listener){
                 listenerToDelete = item;
+                boolean success = listenerToDelete.invalidateUserListener();
+                assert success; // failed if listener was already invalidated
```
ash-mega: Assertions are unreliable on Android. `itr.remove();` will execute even when failed, is it expected?

sergiohs84: Assertions are unreliable?? what does it mean?
Assertions are conditions that are only checked in debugging mode. Every developer must code using debugging mode, so any failed assertion stops the execution and you can check why it failed.
However, in release mode (like it runs in production, regardless logs are enabled or not), the assertion will not be checked at all and therefore the `itr.remove()` will execute.

sergiohs84: What do you mean by _Assertions are unreliable_?

Assertions are conditions that are only checked in debugging mode. Every developer should usually code using debugging mode, so any failed assertion stops the execution and the developer can check why it failed.
However, in release mode (like it runs in production, regardless logs are enabled or not), the assertion will not be checked at all and therefore the `itr.remove()` will execute, as it should be :)

ash-mega: Assertions are unreliable in Dalvik and unimplemented in ART. Should use `BuildConfig.DEBUG` instead.
But only for debugging, it doesn't matter.

ash-mega:
Have tested with Android client, works fine!

PR #755, opened 2019-12-27
#14130. Improve call reconnection procedure

alber2510:
- Remove timers from map although the call was destroyed
- Notify to call handler that reconnection process has finished unsuccessful


sergiohs84:
trigger compilation

src/rtcModule/webrtc.cpp
```diff
@@ -133,13 +133,20 @@ void RtcModule::initInputDevices()
     }
 }
 
-void RtcModule::removeCallRetry(karere::Id chatid)
+void RtcModule::removeCallRetry(karere::Id chatid, bool failed)
```
sergiohs84: Perhaps it's easier to understand if we refactorize this variable into `retry` (being the opposite value than now).

src/rtcModule/webrtc.cpp
```diff
@@ -1959,18 +1966,25 @@ void Call::destroyIfNoSessionsOrRetries(TermCode reason)
...
-        if (wptr.deleted() || mManager.mRetryCall.find(chatid) == mManager.mRetryCall.end())
+        if (wptrManager.deleted())
+        {
             return;
+        }
 
         mManager.mRetryCall.erase(chatid);
         mManager.mRetryCallTimers.erase(chatid);
 
+        if (wptr.deleted() || mManager.mRetryCall.find(chatid) == mManager.mRetryCall.end())
```
sergiohs84: The item in the `mRetryCall` has already been erased, so no point to check its existence
```suggestion
        if (wptr.deleted())
```

src/rtcModule/webrtc.cpp
alber2510: Fixed it

src/rtcModule/webrtcPrivate.h
```diff
@@ -385,7 +385,8 @@ class RtcModule: public IRtcModule, public chatd::IRtcHandler
     //=== Implementation methods
     void initInputDevices();
 
-    void removeCallRetry(karere::Id chatid);
+    void updateConstraints(Resolution resolution);
```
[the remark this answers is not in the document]
alber2510: Fixed it

alber2510:
trigger compilation

PR #757, opened 2020-01-03
#9713. Lean mode to skip fetchnodes

sergiohs84:
In order to allow the app to skip the fetchnodes (and consequently save some memory by not loading all the account's state in memory), this lean mode does not require the fetchnodes to connect to chatd/presenced.


src/megachatapi_impl.cpp
```diff
@@ -1821,6 +1821,12 @@ int MegaChatApiImpl::init(const char *sid)
         localLogout();
     }
 
+    if (!waitForFetchnodesToConnect && state != karere::Client::kInitHasOfflineSession)
+    {
+        localLogout();
+        state = karere::Client::kInitErrGeneric;
```
jnavarrom: Better to use a `setState(state)` here

sergiohs84: Actually, the best would be to manage the error internally by `karere::Client`. I've pushed a commit to do so. It does not affect the logic from the app's point of view :)

javitrujillo:
Based on my tests, I would say this branch is ready and working as expected.

PR #763, opened 2020-01-15
reviewpublicchat file output

cb-mega:
- Ensure we write output file to the exe's path for reviewpublicchat (needed when user double-clicks on exe)
- Ensure we flush every line to the file immediately


mattw-mega:
the `std::` additions affect around half the lines in the file (which will be a hassle for looking up change history), and make it much more bulky to read - is that really needed?  It was compiling fine on all platforms before.

cb-mega:
@mattw-mega Reverted the std:: changes as it actually compiles fine. Please review again.

PR #767, opened 2020-01-20
#14274: Fix try to connect to empty IP's in chatd & presenced

jgandres:


src/net/websocketsIO.cpp
```diff
@@ -309,7 +309,9 @@ bool DNScache::getIp(int shard, std::string &ipv4, std::string &ipv6)
     assert(it->second.mUrl.isValid());
     ipv4 = it->second.ipv4;
     ipv6 = it->second.ipv6;
-    return true;
+
+    // If both Ip's are empty there's no cached ip's
+    return !(ipv4.empty() && ipv6.empty());
```
sergiohs84:
```suggestion
    return ipv4.size() || ipv6.size();
```

PR #768, opened 2020-01-21
#12680. Implement functionality for RTCMD_END_ICE_CANDIDATES

alber2510:
Test Case:
 - Check compatibility with Edge

### DIFF
--- a/src/rtcModule/webrtc.cpp
+++ b/src/rtcModule/webrtc.cpp
@@ -2751,6 +2751,9 @@ void Session::handleMessage(RtMessage& packet)
         case RTCMD_SDP_ANSWER_RENEGOTIATE:
             msgSdpAnswerRenegotiate(packet);
             return;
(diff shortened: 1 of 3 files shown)


src/rtcModule/webrtc.cpp
```diff
@@ -2751,6 +2751,9 @@ void Session::handleMessage(RtMessage& packet)
         case RTCMD_SDP_ANSWER_RENEGOTIATE:
             msgSdpAnswerRenegotiate(packet);
             return;
+        case RTCMD_END_ICE_CANDIDATES:
+            msgSdpAnswerRenegotiate(packet);
```
sergiohs84:
```suggestion
            msgEndIceCandidates(packet);
```
And we should test the reception of this packet does not enter in an endless loop, where an ICE candidate is added, it triggers the `onIceCandidate(null)` callback, it sends a new `END_ICE_CANDIDATES` and so on

PR #770, opened 2020-01-22
Use rtc::scoped_refptr to manage VideoManager

alber2510:
We have change the way of manage a VideoManager object. Now, we use a rtc::scoped_refptr. 
During tests, we detect that changing  the video source during a call produce a memory invalid access.
Although, we have to define a rtc::scoped_refptr in webrtcPrivate.h this solution is less complicated than keep the shared_ptr


javisanesp:
Working on iOS again, thank you @alber2510

javisanesp:
Retested and working 👍

PR #771, opened 2020-01-22
#13760. User attributes outdated for Non-Contacts and Ex-contacts

sergiohs84:
Karere client needs to manage the transition from ex-contact to contact in order to update certain critical attributes (first/lastname). The code has heavily been simplified for better readability.


jgandres:
trigger compilation

PR #777, opened 2020-01-29
#13262. Refactor session notifications

alber2510:
Do not merge this PR until Android and iOS are ready to merge the corresponding integrations.

 - Create new callback at MegaChatApi level onChatSessionUpdate
 - Notify all session changes with this new callback
 - Detail peerid and clientid at notification CHANGE_TYPE_CALL_COMPOSITION
 - Update QT app

### DIFF
--- a/src/megachatapi.h
+++ b/src/megachatapi.h
@@ -84,6 +84,17 @@ public:
         SESSION_STATUS_DESTROYED            /// Session is finished and resources can be released
     };
 
+    enum
+    {
+        CHANGE_TYPE_NO_CHANGES = 0x00,              /// Session doesn't have any change
+        CHANGE_TYPE_STATUS = 0x01,                  /// Session status has changed
+        CHANGE_TYPE_REMOTE_AVFLAGS = 0x02,          /// Remote audio/video flags has changed
+        CHANGE_TYPE_SESSION_NETWORK_QUALITY = 0x04, /// Session network quality has changed
+        CHANGE_TYPE_SESSION_AUDIO_LEVEL = 0x08,    /// Session audio level has changed
+        CHANGE_TYPE_SESSION_OPERATIVE = 0x10,      /// Session is fully operative
+    };
+
+
     virtual ~MegaChatSession();
 
     /**
@@ -139,6 +150,42 @@ public:
     virtual bool hasVideo() const;
 
     /**
+     * @brief Returns the termination code for this session
+     *
+     * Possible values are:
+     *
+     *   MegaChatCall::TERM_CODE_USER_HANGUP       = 0,    /// Normal user hangup
+     *   MegaChatCall::TERM_CODE_CALL_REQ_CANCEL   = 1,    /// Call request was canceled before call was answered
+     *   MegaChatCall::TERM_CODE_CALL_REJECT       = 2,    /// Outgoing call has been rejected by the peer OR incoming call has been rejected in
+                                            /// the current device
+     *   MegaChatCall::TERM_CODE_ANSWER_ELSE_WHERE = 3,    /// Call was answered on another device of ours
+     *   MegaChatCall::TEMR_CODE_REJECT_ELSE_WHERE = 4,    /// Call was rejected on another device of ours
+     *   MegaChatCall::TERM_CODE_ANSWER_TIMEOUT    = 5,    /// Call was not answered in a timely manner
+     *   MegaChatCall::TERM_CODE_RING_OUT_TIMEOUT  = 6,    /// We have sent a call request but no RINGING received within this timeout - no other
+                                            /// users are online
+     *   MegaChatCall::TERM_CODE_APP_TERMINATING   = 7,    /// The application is terminating
+     *   MegaChatCall::TERM_CODE_BUSY              = 9,    /// Peer is in another call
+     *   MegaChatCall::TERM_CODE_NOT_FINISHED      = 10,   /// The call is in progress, no termination code yet
+     *   MegaChatCall::TERM_CODE_DESTROY_BY_COLLISION   = 19,   /// The call has finished by a call collision
+     *   MegaChatCall::TERM_CODE_ERROR             = 21    /// Notify any error type
+     *
+     * @note If the session is not finished yet, it returns MegaChatCall::TERM_CODE_NOT_FINISHED.
+     *
+     * To check if the call was terminated locally or remotely, see MegaChatSession::isLocalTermCode().
+     *
+     * @return termination code for the call
+     */
+    virtual int getTermCode() const;
+
+    /**
+     * @brief Returns if the session finished locally or remotely
+     *
+     * @return True if the call finished locally. False if the call finished remotely
+     */
+    virtual bool isLocalTermCode() const;
+
+
+    /**
      * @brief Returns network quality
      *
      * The valid network quality values are between 0 and 5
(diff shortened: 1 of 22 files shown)


src/megachatapi.h
```diff
@@ -157,6 +204,64 @@ class MegaChatSession
      * @return true if audio is detected for this session, false in other case
      */
     virtual bool getAudioDetected() const;
+
+    /**
+     * @brief Returns a bit field with the changes of the session
+     *
+     * This value is only useful for session notified by MegaChatCallListener::onChatSessionUpdate
+     * that can notify about session modifications. The value only will be valid inside
```
sergiohs84: This value is only useful for session notified by MegaChatCallListener::onChatSessionUpdate ~that can notify about session modifications~.

src/megachatapi.h
```diff
@@ -196,22 +301,17 @@ class MegaChatCall
...
-        CHANGE_TYPE_REMOTE_AVFLAGS = 0x04,          /// Remote audio/video flags has changed
-        CHANGE_TYPE_TEMPORARY_ERROR = 0x08,         /// New temporary error is notified
-        CHANGE_TYPE_RINGING_STATUS = 0x10,          /// Peer has change its ringing state
-        CHANGE_TYPE_SESSION_STATUS = 0x20,          /// Session status has changed
-        CHANGE_TYPE_CALL_COMPOSITION = 0x40,        /// Call composition has changed (User added or removed from call)
-        CHANGE_TYPE_SESSION_NETWORK_QUALITY = 0x80, /// Session network quality has changed
-        CHANGE_TYPE_SESSION_AUDIO_LEVEL = 0x100,    /// Session audio level has changed
-        CHANGE_TYPE_SESSION_OPERATIVE = 0x200      /// Session is fully operative
+        CHANGE_TYPE_TEMPORARY_ERROR = 0x04,         /// New temporary error is notified
+        CHANGE_TYPE_RINGING_STATUS = 0x08,          /// Peer has change its ringing state
```
```suggestion
        CHANGE_TYPE_RINGING_STATUS = 0x08,          /// Peer has changed its ringing state
```

src/megachatapi.h
```diff
@@ -138,6 +149,42 @@ class MegaChatSession
      */
     virtual bool hasVideo() const;
 
+    /**
+     * @brief Returns the termination code for this session
```
sergiohs84: Maybe, based on all those termcodes, it's going to be hard for app's devs to understand which cases the user has left the call vs. cases where there user has drop :thinking:

src/megachatapi_impl.h
```diff
@@ -264,14 +272,15 @@ class MegaChatCallPrivate : public MegaChatCall
...
     int64_t initialTs;
     int64_t finalTs;
     std::string temporaryError;
     std::map<chatd::EndpointId, MegaChatSession *> sessions;
     std::map<chatd::EndpointId, karere::AvFlags> participants;
-    MegaChatHandle peerId;  // to identify the updated session
-    uint32_t clientid;  // to identify the updated session
+    MegaChatHandle peerId;  // to identify participant added or removed
+    uint32_t clientid;  // to identify the participant added or removed
+    bool addedOrRemoved = false;
```
sergiohs84: This kind of booleans should be avoided. They are not easy to understand. Based on the name of the variable, a developer should be able to guess the meaning. `if (addedOrRemoved)` doesn't result in a good readability

src/megachatapi_impl.cpp
```diff
@@ -5136,21 +5182,35 @@ uint8_t MegaChatSessionPrivate::convertSessionState(uint8_t state)
 void MegaChatSessionPrivate::setState(uint8_t state)
 {
     this->state = convertSessionState(state);
+    changed |= MegaChatSession::CHANGE_TYPE_STATUS;
 }
 
 void MegaChatSessionPrivate::setAvFlags(AvFlags flags)
 {
     av = flags;
+    changed |= MegaChatSession::CHANGE_TYPE_REMOTE_AVFLAGS;
```
sergiohs84: Double-check this change is not reported for changes on our own AV flags, but only for remote (and viceversa)

src/megachatapi.h
```diff
@@ -501,24 +593,34 @@ class MegaChatCall
...
      *
-     * This function only returns a valid value when session status change is notified
+     * This function only returns a valid value when MegaChatCall::CHANGE_TYPE_CALL_COMPOSITION is notified
      * via MegaChatCallListener::onChatCallUpdate
      *
-     * @return Handle of the client which session has changed its status
+     * @return true/false if user with peerid-clientid has been added/removed from call
      */
-    virtual MegaChatHandle getClientidSessionStatusChange() const;
+    virtual bool getClientIsAddedOrRemoved() const;
```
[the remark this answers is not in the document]
alber2510: In documentation I set:
`This function only returns a valid value when MegaChatCall::CHANGE_TYPE_CALL_COMPOSITION is notified via MegaChatCallListener::onChatCallUpdate`
But it's true that can be confusing for app developers

src/megachatapi_impl.cpp
```diff
@@ -5049,8 +5068,15 @@ MegaChatSessionPrivate::MegaChatSessionPrivate(const rtcModule::ISession &sessio
 }
 
 MegaChatSessionPrivate::MegaChatSessionPrivate(const MegaChatSessionPrivate &session)
-    : state(session.getStatus()), peerid(session.getPeerid()), clientid(session.getClientid()), av(session.hasAudio(), session.hasVideo()),
-      networkQuality(session.getNetworkQuality()), audioDetected(session.getAudioDetected())
+    : state(session.getStatus())
+    , peerid(session.getPeerid())
```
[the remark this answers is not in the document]
alber2510: I've added the comma at beginning to avoid that if we add a new member initialization the PR diff code shows changes in two lines, similar to enumeration with adding the last comma

src/megachatapi_impl.cpp
```diff
@@ -5648,18 +5707,15 @@ bool MegaChatCallPrivate::addOrUpdateParticipant(Id userid, uint32_t clientid, A
     std::map<chatd::EndpointId, karere::AvFlags>::iterator it = participants.find(endPointId);
     if (it == participants.end())   // new participant
     {
-        changed |= MegaChatCall::CHANGE_TYPE_CALL_COMPOSITION;
+        this->changed |= MegaChatCall::CHANGE_TYPE_CALL_COMPOSITION;
```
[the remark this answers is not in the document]
sergiohs84: Let's keep it as designed by now, but consider to create more specific lightweight callbacks in the future for other purposes.

src/megachatapi_impl.cpp
```diff
@@ -8059,8 +8112,16 @@ bool MegaChatCallHandler::isParticipating(Id userid)
 
 void MegaChatCallHandler::removeAllParticipants()
 {
-    chatCall->removeAllParticipants();
-    megaChatApi->fireOnChatCallUpdate(chatCall);
+    MegaHandleList* clientids = chatCall->getClientidParticipants();
```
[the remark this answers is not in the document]
sergiohs84: Alright, let's make clear for apps devs that, upon a reconnecting status, these callbacks can and should be ignored (the participants have not left/drop the call, but you did).
Additionally, check that the addition of the existing participants to the call during the transition from _Reconnecting_ to _In progress_ happens before the new status _In progress_ is notified (so the apps also ignore these callbacks until the call is fully established again). It should happen like that already, since upon reconnection the calldata of each current participant is received before the own user starts to re-join/reconnect the call.

src/megachatapi.h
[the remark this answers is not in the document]
sergiohs84: Let's move to the `getCallCompositionChange()` approach.

src/megachatapi.h
```diff
@@ -138,6 +149,42 @@ class MegaChatSession
...
+                                            /// the current device
+     *   MegaChatCall::TERM_CODE_ANSWER_ELSE_WHERE = 3,    /// Call was answered on another device of ours
+     *   MegaChatCall::TEMR_CODE_REJECT_ELSE_WHERE = 4,    /// Call was rejected on another device of ours
+     *   MegaChatCall::TERM_CODE_ANSWER_TIMEOUT    = 5,    /// Call was not answered in a timely manner
+     *   MegaChatCall::TERM_CODE_RING_OUT_TIMEOUT  = 6,    /// We have sent a call request but no RINGING received within this timeout - no other
+                                            /// users are online
+     *   MegaChatCall::TERM_CODE_APP_TERMINATING   = 7,    /// The application is terminating
+     *   MegaChatCall::TERM_CODE_BUSY              = 9,    /// Peer is in another call
+     *   MegaChatCall::TERM_CODE_NOT_FINISHED      = 10,   /// The call is in progress, no termination code yet
+     *   MegaChatCall::TERM_CODE_DESTROY_BY_COLLISION   = 19,   /// The call has finished by a call collision
```
sergiohs84: Check the use-case where there is a collision at session level, in order to decide if the app needs to know about the session being destroyed by this situation.

Additionally, we should remove all the values that are not possible at session-level (only at call-level). And adjust the comment:
```suggestion
     - MegaChatCall::TERM_CODE_DESTROY_BY_COLLISION   = 19
      The session has finished by a session collision
```

src/megachatapi.h
sergiohs84: Simply indicate in the explanation of the codes whether it means the user has left, or the user has dropped.

src/megachatapi_impl.cpp
sergiohs84: Checked.

src/megachatapi.h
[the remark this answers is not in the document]
alber2510: Fixed

src/megachatapi_impl.cpp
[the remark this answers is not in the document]
alber2510: Added explanation to documentation

src/megachatapi_impl.h
alber2510: Fixed

src/megachatapi.h
alber2510: Fxied

src/megachatapi.h
```diff
@@ -154,22 +154,17 @@ class MegaChatSession
...
+     *  Normal user hangup. User has left the call
+     *
+     *   - MegaChatCall::TERM_CODE_NOT_FINISHED      = 10
+     *  The session is in progress, no termination code yet
+     *
+     *   -MegaChatCall::TERM_CODE_ERROR             = 21
+     *  Notify any error type. A reconnection is launched
      *
      * @note If the session is not finished yet, it returns MegaChatCall::TERM_CODE_NOT_FINISHED.
+     * The rest of values are invalid as term code  for a session
```
```suggestion
     * The rest of values are invalid as term code for a session
```

sergiohs84:
trigger compilation

sergiohs84:
trigger compilation

sergiohs84:
trigger compilation

sergiohs84:
trigger compilation

src/rtcModule/webrtc.cpp
```diff
@@ -782,7 +783,13 @@ void RtcModule::removeCall(Id chatid, bool retry)
...
             if (retry || itHandler->second->callParticipants())
             {
-                itHandler->second->removeAllParticipants();
+                bool reconnectionState = false;
+                if (mRetryCall.find(chatid) != mRetryCall.end())
+                {
+                    reconnectionState = true;
+                }
+
+                itHandler->second->removeAllParticipants(reconnectionState);
```
sergiohs84, line 792: Next time, please, try to simply do:
```suggestion
                itHandler->second->removeAllParticipants(mRetryCall.find(chatid) != mRetryCall.end());
```

PR #781, opened 2020-02-03
#14448. Release video device when call is finished

alber2510:
Android team has detected an issue with mobile `MI 9T`. The camera doesn't close after a call


javisanesp:
Just tested with @alber2510, nothing crashed 🙂

alber2510:
Tested in Android and works fine too

PR #783, opened 2020-02-06
Update setup-deps.sh

kirssomas:


sergiohs84:
Please, @kirssomas , can you explain the purpose of your change?
It is syntactically incorrect, and you only add a MEGA link to a `.rar` file about how to make videos from YouTube :thinking:

PR #785, opened 2020-02-10
#14470. When I change my email, it doesn't update in megaChat

jgandres:


src/chatClient.cpp
```diff
@@ -3976,6 +3984,19 @@ std::string Client::getUserAlias(uint64_t userId)
     return aliasBin;
 }
 
+void Client::setMyEmail(const char *email)
+{
+    if (email && email[0])
+    {
+        mMyEmail.assign(email);
+    }
```
sergiohs84:
```
else
{
    mMyEmail.clear();
}
```

src/chatClient.cpp
```diff
@@ -3645,6 +3645,14 @@ void ContactList::syncWithApi(mega::MegaUser& user)
         {
             contact->mEmail = user.getEmail();
             client.db.query("update contacts set email = ? where userid = ?", contact->email(), handle);
+            client.userAttrCache().onUserAttrChange(user);
```
sergiohs84: It would be good to avoid re-fetch the email by sending a new API command. Instead, you could update the value here immediately, from the information received in the `MegaUser`. Only if it's relatively simple, otherwise email changes doesn't happen very often, so it would be fine.

jgandres:
trigger compilation

src/chatClient.cpp
```diff
@@ -3664,10 +3678,23 @@ void ContactList::syncWithApi(mega::MegaUser& user)
...
-        client.userAttrCache().onUserAttrChange(userid, changed);
+        changed = ::mega::MegaUser::CHANGE_TYPE_FIRSTNAME | ::mega::MegaUser::CHANGE_TYPE_LASTNAME | ::mega::MegaUser::CHANGE_TYPE_EMAIL;
+    }
+
+    if (user.getChanges() && !user.isOwnChange())
+    {
+        // update user attributes if changes are external
+        client.userAttrCache().onUserAttrChange(user);
+    }
+    else if (changed)
```
sergiohs84: Isn't `user.getChanges() == changed)`?? If so, why do we need the if-else?

sergiohs84: Already revamped all the logic around `changed`

src/chatClient.cpp
sergiohs84: It would require to tweak the update of the `userAttrCache`, not clean. Let's fetch a fresh email in these cases.

PR #789, opened 2020-02-18
#14598. Fix invalid access to variable after destroy the object

alber2510:
The variable mManager has invalid value when Call object is destroyed. We keep mManager pointer to use in the timer

### DIFF
--- a/src/rtcModule/webrtc.cpp
+++ b/src/rtcModule/webrtc.cpp
@@ -1976,16 +1976,17 @@ void Call::destroyIfNoSessionsOrRetries(TermCode reason)
     }
 
     auto wptr = weakHandle();
-    auto wptrManager = mManager.weakHandle();
-    mManager.mRetryCallTimers[chatid] = setTimeout([this, wptr, wptrManager, chatid, reason]()
+    RtcModule* manager = &mManager;
+    auto wptrManager = manager->weakHandle();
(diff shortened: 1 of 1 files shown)


src/rtcModule/webrtc.cpp
```diff
@@ -1976,16 +1976,17 @@ void Call::destroyIfNoSessionsOrRetries(TermCode reason)
     }
 
     auto wptr = weakHandle();
-    auto wptrManager = mManager.weakHandle();
-    mManager.mRetryCallTimers[chatid] = setTimeout([this, wptr, wptrManager, chatid, reason]()
+    RtcModule* manager = &mManager;
+    auto wptrManager = manager->weakHandle();
+    mManager.mRetryCallTimers[chatid] = setTimeout([this, wptr, wptrManager, manager, chatid, reason]()
```
sergiohs84: Should we additionally specify the copy-ctor as deleted for `RtcModule`? That way, this issue would be caught at compile time.

alber2510:
It's necessary a hotfix, we create a new branch from realeas

src/rtcModule/webrtc.cpp
sergiohs84: Nevermind, I see now that your aim was to keep the pointer.

PR #790, opened 2020-02-18
#14602. Code maintenance: clean appctx

sergiohs84:
The `appCtx` is used everywhere in MEGAchat, since it's required by `marshallCall()`. I makes it easier if the `appCtx` is not passed all the way to wherever it's used, but initialized at startup (`MegaApiImpl` ctor) and used exclusively where it's mandatory, inside the `marshallCall()` in order to post the message for `MegaApiImpl` to queue the job.
This PR also deletes an obsolete file: `asyncTools`


sergiohs84:
This approach only works when there is a single application context. As soon as there are more than one `MegaChatApi` instances, it doesn't work.
Closing...

PR #797, opened 2020-02-21
#14653. Improve app behavior at session logout during call

alber2510:


src/chatClient.cpp
```diff
@@ -1504,11 +1504,6 @@ void Client::onConnStateChange(presenced::Client::ConnState /*state*/)
 
 void Client::terminate(bool deleteDb)
 {
-#ifndef KARERE_DISABLE_WEBRTC
```
sergiohs84: I think we should keep this call here, since the `terminate()` can be call for different reasons than a logout, and in all cases the calls should be hang up.

src/megachatapi_impl.cpp
```diff
@@ -341,9 +341,7 @@ void MegaChatApiImpl::sendPendingRequests()
         case MegaChatRequest::TYPE_LOGOUT:
         {
             bool deleteDb = request->getFlag();
-#ifndef KARERE_DISABLE_WEBRTC
```
sergiohs84: This `#ifndef` should be removed also from the `MegaChatRequest::TYPE_DELETE` case.

src/megachatapi_impl.cpp
```diff
@@ -7907,7 +7909,12 @@ void MegaChatCallHandler::onDestroy(rtcModule::TermCode reason, bool /*byPeer*/,
     {
         chatid = chatCall->getChatid();
         unique_ptr<MegaChatRoom> chatRoom(megaChatApi->getChatRoom(chatid));
-        assert(chatRoom);
+        if (!chatRoom)
+        {
+            // ChatRoom has been deleted => logout
```
sergiohs84: I would keep an `assert(false)` here, since it represents a bug on the cleanup phase.

PR #798, opened 2020-02-21
#14656.  Call gets in reconnected state after answer

alber2510:
In some situations, IOS app is forcing a network reconnection after answer a call. In normal condition, call reconnection is successful but when we have some network issue, the reconnection finishes with failure (after time-out) and call is finished unsuccessfully
This PR solve the issue and call reconnection will be successful if app recovers the network correctly


sergiohs84:
trigger compilation

sergiohs84:
trigger compilation

PR #805, opened 2020-03-05
#14685. Chat link scalability

alber2510:
Wait for iOS and Android to adjust the UI before merge this PR


src/chatClient.h
```diff
@@ -280,6 +281,7 @@ class PeerChatRoom: public ChatRoom
     void initContact(const uint64_t& peer);
     void updateChatRoomTitle();
 
+    virtual bool isMember(karere::Id peerid) const override;
```
sergiohs84: `virtual` is not needed when `override`. I'll adjust it

src/chatClient.h
```diff
@@ -195,6 +195,8 @@ class ChatRoom: public chatd::Listener, public DeleteTrackable
 
     bool hasChatHandler() const;
 
+    virtual unsigned long numMembers() const = 0;
```
sergiohs84: I think it should count your own user as participants, so `2` for 1on1 and `mPeers.size() + 1` for groupchats.

src/megachatapi_impl.cpp
```diff
@@ -1426,6 +1435,13 @@ void MegaChatApiImpl::sendPendingRequests()
                 break;
             }
 
+            if (chatroom->publicChat() && chatroom->numMembers() > MAX_PUBLICCHAT_MEMBERS_FOR_CALL)
+            {
+                API_LOG_ERROR("Start call - the public chat has too many participants");
+                errorCode = MegaChatError::ERROR_ACCESS;
```
sergiohs84: Why not `ETOOMANY` for calls, similar than to make-private? (here and also at `ANSWER_CALL`

src/userAttrCache.h
```diff
@@ -175,7 +175,7 @@ class UserAttrCache: public std::map<UserAttrPair, std::shared_ptr<UserAttrCache
      * every time the attribute changes on the server and the new value is fetched.
      */
     Handle getAttr(uint64_t user, unsigned attrType, void* userp,
-                             UserAttrReqCbFunc cb, bool oneShot=false, uint64_t ph = Id::inval());
+                             UserAttrReqCbFunc cb, bool oneShot=false, bool fetch = true, uint64_t ph = Id::inval());
```
sergiohs84: I miss the documentation of the new param `fetch`

src/chatd.cpp
```diff
@@ -2947,6 +2947,16 @@ void Chat::attachmentHistDone()
     mAttachmentNodes->finishFetchingFromServer();
 }
 
+void Chat::requestUserAttributes(Id sender)
+{
+    if (!isGroup() && !isPublic())
```
sergiohs84: Not sure, but I think you want here:
```suggestion
    if (!(isGroup() && isPublic()))
```
If so, you can directly skip the `isGroup()` (all public chats are groupchats).
```suggestion
    if (!isPublic())
```

src/chatd.cpp
```diff
@@ -4672,6 +4672,11 @@ void Chat::msgIncomingAfterDecrypt(bool isNew, bool isLocal, Message& msg, Idx i
         CALL_LISTENER(onMsgOrderVerificationFail, msg, idx, "A message with that backrefId "+std::to_string(msg.backRefId)+" already exists");
     }
 
+    if (isPublic() && msg.userid != mChatdClient.mMyHandle)
```
sergiohs84: If there is a way of checking the attributes are already fetched, perhaps we can skip the request, considering this piece of code will be executed per message.

alber2510: I think the request would be generated the first time only. 
In method `UserAttrCache::getAttr(uint64_t userHandle, unsigned type, void* userp, UserAttrReqCbFunc cb, bool oneShot, bool fetch, uint64_t ph)` we would have an element in the map and item.pending would be different form kCacheFetchNewPending and kCacheNotFetchUntilUse, then we call directly to the callback.
Although request is not launched, we make many unnecessary method calls. Implement a method to check if fetch is necessary it would be ok

src/chatClient.h
alber2510: Fixed

src/megachatapi_impl.cpp

src/userAttrCache.h
alber2510: Fixed

src/chatd.cpp
alber2510: Changed

sergiohs84:
trigger compilation

src/megachatapi_impl.cpp
```diff
@@ -2679,11 +2719,11 @@ void MegaChatApiImpl::getUserEmail(MegaChatHandle userhandle, MegaChatRequestLis
     waiter->notify();
 }
 
-void MegaChatApiImpl::loadUserAttributes(MegaChatHandle chatid, MegaHandleList userList, const char *authorizationToken, MegaChatRequestListener *listener)
+void MegaChatApiImpl::loadUserAttributes(MegaChatHandle chatid, MegaHandleList *userList, const char *authorizationToken, MegaChatRequestListener *listener)
 {
     MegaChatRequestPrivate *request = new MegaChatRequestPrivate(MegaChatRequest::TYPE_GET_PEER_ATTRIBUTES, listener);
     request->setChatHandle(chatid);
-    request->setMegaHandleList(userList.copy());
+    request->setMegaHandleList(userList->copy());
```
sergiohs84, line 2726: You are not testing `userList` (it could be null). Apart from that, the `setMegaHandleList()` already does a copy, so it's a waste to pass a copy here.

src/megachatapi_impl.cpp
```diff
@@ -1855,13 +1855,10 @@ void MegaChatApiImpl::sendPendingRequests()
...
             }
 
-            const char* publicHandle = request->getLink();
-            MegaChatHandle ph = publicHandle ? karere::Id(publicHandle, strlen(publicHandle)).val : MEGACHAT_INVALID_HANDLE;
-
             std::vector<::promise::Promise<void>> promises;
             for (unsigned int i = 0; i < handleList->size(); i++)
             {
-                promises.push_back(mClient->userAttrCache().getAttributes(handleList->get(i), ph));
+                promises.push_back(chatroom->chat().requestUserAttributes(handleList->get(i)));
```
alber2510, line 1861: Maybe the request can finish with an MegaChatError if we call this method with a private chatroom. The method `requestUserAttributes` returns a promise::_Void() and doesn't do anything. For private chatrooms, it isn't usefull

examples/megaclc/megaclc.cpp
```diff
@@ -1159,11 +1159,13 @@ void reportMessageHuman(c::MegaChatHandle chatid, c::MegaChatMessage *msg, const
         const char* firstname_ptr = nullptr;
         if (room)
         {
-            firstname_ptr = room->getPeerFirstnameByHandle(handle);
+            firstname_ptr = g_chatApi->getUserFirstnameFromCache(handle);
```
sergiohs84: unique_ptr

examples/megaclc/megaclc.cpp
```diff
@@ -1159,11 +1159,13 @@ void reportMessageHuman(c::MegaChatHandle chatid, c::MegaChatMessage *msg, const
...
         }
         if (firstname_ptr && *firstname_ptr != '\0')
         {
-            return std::string{firstname_ptr};
+            std::string name(firstname_ptr);
+            delete [] firstname_ptr;
+            return name;
```
sergiohs84: Simply return `firstname_ptr->get()` and it will implicitly create the `string` from the `const char *`. Remember it's a C-string ;)

examples/megaclc/megaclc.cpp
```diff
@@ -1185,11 +1187,13 @@ void reportMessageHuman(c::MegaChatHandle chatid, c::MegaChatMessage *msg, const
         const char* lastname_ptr = nullptr;
         if (room)
         {
-            lastname_ptr = room->getPeerLastnameByHandle(handle);
+            lastname_ptr = g_chatApi->getUserLastnameFromCache(handle);
```
sergiohs84: unique_ptr

src/chatClient.cpp
```diff
@@ -3546,23 +3546,43 @@ void GroupChatRoom::enablePreview(uint64_t ph)
 
 bool GroupChatRoom::publicChat() const
 {
-    return (mChat->crypto()->isPublicChat());
+    if (mChat)
```
sergiohs84, line 3549: Is it not guaranteed that this method is called only once the `mChat` is initialized?? Otherwise, there should be a bug, and this change (along with others in `getPublicHandle()`, `getNumPreviewers()` and `previewMode()` may have unexpected consequences, since the returned value will be affected by race conditions. @alber2510 , can you explain, please?

src/megachatapi_impl.cpp
sergiohs84: This comment is obsolete, right?

src/megachatapi_impl.cpp
```diff
@@ -2747,12 +2747,7 @@ const char *MegaChatApiImpl::getUserFirstnameFromCache(MegaChatHandle userhandle
     SdkMutexGuard g(sdkMutex);
     if (mClient && mClient->isUserAttrCacheReady())
     {
-        const Buffer* buffer = mClient->userAttrCache().getDataFromCache(userhandle, ::mega::MegaApi::USER_ATTR_FIRSTNAME);
-        if (buffer != nullptr)
-        {
-            std::string name(buffer->buf(), buffer->size());
-            return MegaApi::strdup(name.c_str());
-        }
+        return move(mClient->userAttrCache().getDataFromCache(userhandle, ::mega::MegaApi::USER_ATTR_FIRSTNAME)->c_str());
```
alber2510, line 2750: I think `move` method isn't necessary, we have a raw pointer and we are returning it

sergiohs84: Agree, change it, please.

src/chatClient.cpp
```diff
@@ -3546,43 +3546,38 @@ void GroupChatRoom::enablePreview(uint64_t ph)
...
 
 uint64_t GroupChatRoom::getPublicHandle() const
 {
+    assert(mChat);
     if (mChat)
     {
         return (mChat->getPublicHandle());
     }
 
+    parent.mKarereClient.api.callIgnoreResult(&::mega::MegaApi::sendEvent, 99011, "GroupChatRoom::getPublicHandle(), chatd::Chat isn't yet created");
```
sergiohs84, line 3568: For the same event id, the stats server only records the name of the first occurrence. In result, you will not be able to differentiate whether the event comes from `getPublicHandle()` or from `publicChat()`. I think that using only one event is enough, anyway :)

src/chatClient.cpp
sergiohs84: We assume that the lambda using this method at the `Member` will not call it before `mChat` is initialized, but just in case we decided to avoid a potential race and avoid the segfault, but notify MEGA with an event.

src/megachatapi_impl.cpp
```diff
@@ -2768,6 +2768,17 @@ const char *MegaChatApiImpl::getUserLastnameFromCache(MegaChatHandle userhandle)
...
+const char *MegaChatApiImpl::getUserFullnameFromCache(MegaChatHandle userhandle)
+{
+    SdkMutexGuard g(sdkMutex);
+    if (mClient && mClient->isUserAttrCacheReady())
+    {
+        const Buffer* buffer = mClient->userAttrCache().getDataFromCache(userhandle, USER_ATTR_FULLNAME);
+        if (buffer != nullptr)
+        {
+            std::string fullname(buffer->buf()+1, buffer->dataSize()-1);
+            return MegaApi::strdup(fullname.c_str());
```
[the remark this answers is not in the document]
sergiohs84: Since we give the ownership to the caller, they need the actual pointer returned by `new []`, otherwise it will result on undefined behavior.
In short, disregard my suggestion.

src/chatClient.cpp
sergiohs84: Since the caller takes ownership, they need the same address than `new []` returned. Otherwise `delete []` will result on undefined behavior.
In short, disregard my suggestion. Thanks

src/chatClient.cpp
```diff
@@ -3831,6 +3838,7 @@ bool GroupChatRoom::syncWithApi(const mega::MegaTextChat& chat)
         onArchivedChanged(mIsArchived);
     }
 
+    mAutoJoiningCompleted = true;
```
sergiohs84, line 3841: I think this is not the right place, since it will set this value, in example, when the title changes.
The flag should be set exclusively when the own user has rejoined the room.

src/chatClient.cpp
```diff
@@ -3243,6 +3243,7 @@ promise::Promise<void> GroupChatRoom::invite(uint64_t userid, chatd::Priv priv)
 promise::Promise<void> GroupChatRoom::autojoinPublicChat(uint64_t ph)
 {
     Id myHandle(parent.mKarereClient.myHandle());
+    mAutoJoiningCompleted = false;
```
sergiohs84, line 3246: If the request fails, we need to set the flag back to `true`, right?

src/chatd.cpp
```diff
@@ -5191,6 +5191,8 @@ void Chat::onUserJoin(Id userid, Priv priv)
         CALL_CRYPTO(onUserJoin, userid);
     }
 
+    ///TODO: in public link extra notification can be generated if we are re-invited
+    /// and chatd joins arrive before than api notification
```
sergiohs84, line 5195: I think this is not the right place for the comment (that needs to indicate both, the auto-join from another client and the invitation by a moderator). The right place is where the notifications are filtered out, with some context about why they are skipped and why not in the cases aforementioned.

src/chatClient.cpp

src/chatd.cpp
alber2510: Fixed

src/chatClient.cpp
alber2510: Fixed

src/chatClient.cpp
```diff
@@ -3602,13 +3659,23 @@ bool GroupChatRoom::syncMembers(const mega::MegaTextChat& chat)
         }
     }
 
+    if (commitEach)
+    {
+        parent.mKarereClient.setCommitMode(true);
+    }
```
sergiohs84, line 3665: No need to check previous state. If it has not changed, it does nothing.
```suggestion cpp
        parent.mKarereClient.setCommitMode(commitEach);
```

src/chatd.cpp
```diff
@@ -5196,12 +5213,18 @@ void Chat::onUserLeave(Id userid)
...
         for (auto &it: mUsers)
         {
             CALL_CRYPTO(onUserLeave, it);
             CALL_LISTENER(onUserLeave, it);
         }
         mUsers.clear();
+        if (commitEach)
+        {
+            mChatdClient.mKarereClient->setCommitMode(true);
+        }
```
sergiohs84, line 5227: No need to check previous state. If it has not changed, it does nothing.
```suggestion cpp
            mChatdClient.mKarereClient->setCommitMode(commitEach);
```

src/chatd.cpp
```diff
@@ -5176,8 +5186,15 @@ void Chat::onUserJoin(Id userid, Priv priv)
...
 
     mUsers.insert(userid);
-    CALL_CRYPTO(onUserJoin, userid);
-    CALL_LISTENER(onUserJoin, userid, priv);
+    if (!isPublic())
+    {
+        CALL_CRYPTO(onUserJoin, userid);
+    }
+
+    if (mOwnPrivilege != Priv::PRIV_NOTPRESENT && !isPublic())
```
sergiohs84, line 5194:
```suggestion cpp
    // avoid to notify if own user doesn't participate and it's a public chat (for large chat-links, for performance)
    if (mOwnPrivilege != Priv::PRIV_NOTPRESENT && !isPublic())
```

src/chatClient.cpp
alber2510: Yes, but maybe we want to keep in commit mode false, that is why I'm checking the previous state. The commit mode keep in commit mode false while I'm not connected to all chats. In that point, I don't know if I'm connected to all chats or I have to receive more info from other chats

src/chatClient.cpp
```diff
@@ -3226,11 +3256,16 @@ promise::Promise<void> GroupChatRoom::autojoinPublicChat(uint64_t ph)
         std::string uKeyB64;
         mega::Base64::btoa(uKeyBin, uKeyB64);
 
+        parent.mKarereClient.setCommitMode(false);
```
sergiohs84, line 3259: Not sure, but... should we restore the previous mode before autojoin the public chat? or is it intentionally set to false?

alber2510: It's intentionally set false to avoid make a commit for every new member. We set true again at `GroupChatRoom::syncMembers`

aabellagm:
trigger compilation android

aabellagm:
trigger compilation android

aabellagm:
trigger compilation android

aabellagm:
trigger compilation android

aabellagm:
trigger compilation android

aabellagm:
trigger compilation android

aabellagm:
trigger compilation android

aabellagm:
trigger compilation android

aabellagm:
trigger compilation android

aabellagm:
trigger compilation android

PR #814, opened 2020-03-16
#14949. Set kStateInitial when join to group call, when call is in-progress

alber2510:


alber2510:
It's replaced by this one #815

PR #816, opened 2020-03-17
#14967. Send full name instead alias when attach contact in MegaChat

jgandres:


sergiohs84:
trigger compilation

examples/qtmegachatapi/MegaChatApplication.cpp
```diff
@@ -680,21 +680,31 @@ void MegaChatApplication::onRequestFinish(MegaChatApi *, MegaChatRequest *reques
...
              MegaChatHandle userHandle = request->getUserHandle();
              int errorCode = error;
+             mFirstnameFetching[userHandle] = false;
              if (errorCode == MegaChatError::ERROR_OK)
              {
                 const char *firstname = request->getText();
                 if ((strlen(firstname)) == 0)
                 {
-                    this->mMegaChatApi->getUserEmail(userHandle);
+                    auto it = mFirstnamesMap.find(userHandle);
```
alber2510: It's not necessary `find` in a map before `erase`. `erase` from a map with a non-existent key is safe. The code is correct. You can remove if you want or keep in this way

examples/qtmegachatapi/MegaChatApplication.cpp
```diff
@@ -680,21 +680,31 @@ void MegaChatApplication::onRequestFinish(MegaChatApi *, MegaChatRequest *reques
...
                     break;
                 }
                 mFirstnamesMap[userHandle] = firstname;
-                mFirstnameFetching[userHandle] = false;
                 mMainWin->updateContactTitle(userHandle,firstname);
                 mMainWin->updateMessageFirstname(userHandle,firstname);
              }
              else if (errorCode == MegaChatError::ERROR_NOENT)
              {
+                auto it = mFirstnamesMap.find(userHandle);
```
alber2510: It's not necessary `find` in a map before `erase`. `erase` from a map with a non-existent key is safe. The code is correct. You can remove if you want or keep in this way

jgandres:
trigger compilation

PR #818, opened 2020-03-18
#12205. Initialize user attributes with the 'ug' response

alber2510:
Remove public RSA key and retrieves private RSA directly from SDK


alber2510:
This PR is unnecessary due to change introduced by this one https://github.com/meganz/MEGAchat/pull/842

PR #826, opened 2020-03-25
#14784. Reduce latency to connect a call

alber2510:


alber2510:
trigger compilation

src/rtcModule/webrtc.cpp
```diff
@@ -853,6 +906,51 @@ void RtcModule::abortCallRetry(Id chatid)
...
 }
 
+void RtcModule::refreshTurnServerIp()
+{
+    if (mIceServerProvider.empty())
+    {
+        return;
+    }
+
+    std::string url = mIceServerProvider[0]->url;
```
sergiohs84, line 916: The API may return a list of servers (despite currently it only returns 1). In the past, those servers used to be stored in the `mIceServerProvider` and the old call to `setIceServers(mIceServerProvider)` iterate through all those servers. 

However, now we only do it for the first server in the list and we should be considering all of them.

I can see the problem: you only plan to cache a single URL+IP, regardless the number of servers returned by API / geLB. On top off my head, I can think of multiple URLs in the cache by simply using the `TURNSERVER_SHARD + index` as id, where the index is the position in the list for each URL?. If you see complications on this suggestion, let's discuss about it in a call, @alber2510

src/rtcModule/webrtc.cpp
```diff
@@ -853,6 +906,51 @@ void RtcModule::abortCallRetry(Id chatid)
...
+        std::string subString = url.substr(postInitialColon, postFinalColon - postInitialColon);
+        auto wptr = weakHandle();
+        int turnServerDnsRequestId = ++mTurnServerDnsRequest;
+        DnsRequest::getInstance()->wsResolveDNS(mKarereClient.websocketIO, subString.c_str(),
+                                                [wptr, this, subString, turnServerDnsRequestId]
+                                                (int statusDNS, const std::vector<std::string> &ipsv4, const std::vector<std::string> &ipsv6)
+        {
+            if (wptr.deleted() || turnServerDnsRequestId != mTurnServerDnsRequest)
+                return;
+
```
sergiohs84, line 930: I miss the management of `statusDNS < 0`, which indicates the DNS resolution has failed. In that case, I think the DNS records should not be updated.
Also, it would be good to add some error logging in this lambda function in order to ease debugging.

src/rtcModule/webrtc.cpp
```diff
@@ -853,6 +906,51 @@ void RtcModule::abortCallRetry(Id chatid)
...
+            if (!mKarereClient.mDnsCache.hasRecord(TURNSERVER_SHARE))
```
sergiohs84: Not sure if those records should be already added at this point. The right place to add them to the cache, IMHO, are in the lambda of `fetchServers()`. At that point the URL/s have been received and could be added to the cache.

src/rtcModule/webrtc.cpp
```diff
@@ -853,6 +906,51 @@ void RtcModule::abortCallRetry(Id chatid)
...
+            }
+
+            if (ipsv4.size() || ipsv6.size())
+            {
+                std::vector<std::string> ips;
+                ips.push_back("127.0.0.1");
+                mKarereClient.mDnsCache.setIp(TURNSERVER_SHARE, ipsv4, ipsv6);
+                KR_LOG_DEBUG("New IP for Turn servers: ipv4 - %s     ipv6 - %s", ipsv4.size() ? ipsv4[0].c_str() : "", ipsv6.size() ? ipsv6[0].c_str() : "");
+                StaticProvider iceServer = getTurnServerProvider(ipsv4[0], ipsv6[0]);
+                setIceServers(iceServer);
```
sergiohs84, line 948: Here, probably it's more appropriate to use the `addIceServer()` instead, so we keep the original list plus the new server, instead of replace it.

src/rtcModule/webrtc.cpp
```diff
@@ -1794,7 +1892,7 @@ void Call::removeSession(Session& sess, TermCode reason)
     }
 
     TermCode terminationCode = (TermCode)(reason & ~TermCode::kPeer);
-    if (terminationCode == TermCode::kErrIceFail || terminationCode == TermCode::kErrIceTimeout)
+    if (terminationCode == TermCode::kErrIceFail || terminationCode == TermCode::kErrIceTimeout || terminationCode == kErrSessSetupTimeout)
```
sergiohs84, line 1895: Is it related to this PR?

src/rtcModule/webrtcPrivate.h
```diff
@@ -486,6 +493,46 @@ class RtMessageComposer: public chatd::Command
...
+    bool wsConnect(WebsocketsIO *websocketIO, const char *ip,
+                   const char *host, int port, const char *path, bool ssl) = delete;
+    int wsGetNoNameErrorCode(WebsocketsIO *websocketIO) = delete;
+    bool wsSendMessage(char *msg, size_t len) = delete;  // returns true on success, false if error
+    void wsDisconnect(bool immediate) = delete;
+    bool wsIsConnected() = delete;
+    void wsCloseCbPrivate(int errcode, int errtype, const char *preason, size_t reason_len) = delete;
+
+    void wsConnectCb(){}
+    virtual void wsCloseCb(int errcode, int errtype, const char *preason, size_t /*preason_len*/){}
```
sergiohs84: We should use override in these cases. I'll change it

src/rtcModule/webrtcPrivate.h
```diff
@@ -486,6 +493,46 @@ class RtMessageComposer: public chatd::Command
...
+    {
+        if (!instance)
+        {
+            instance = new DnsRequest();
+        }
+
+        return instance;
+    }
+
+    ~DnsRequest()
```
sergiohs84: Always declare virtual destructors for child classes. I'll do it

src/rtcModule/webrtc.cpp
```diff
@@ -87,15 +89,31 @@ RtcModule::RtcModule(karere::Client& client, IGlobalHandler& handler,
...
 void RtcModule::init()
 {
-    StaticProvider iceServerStatic(mStaticIceSever);
+    string ipv4;
+    string ipv6;
+    StaticProvider iceServerStatic;
+    mKarereClient.mDnsCache.getIp(TURNSERVER_SHARE, ipv4, ipv6);
+    if (ipv4.size() || ipv6.size())
+    {
+        iceServerStatic = getTurnServerProvider(ipv4, ipv6);
```
sergiohs84, line 98: If possible, it would be better to reuse the code to `setServers()`, perhaps by adding the IPs as optional param. It could be not a good idea to prepare the JSON with fixed values. The API returns the service (`turn`), the port (`3478`) and even the protocol (`udp`). Ideally, we should keep using the values returned by the API originally, so in case the API needs to adjust any parameter, we don't need changes.
Maybe the returned JSON could be stored in the URL column in the `dns_cache` DB table. So it can be used here, and if the IPs are also in cache, use then instead of the domain name. What do you think, @alber2510 ??

alber2510: I tried in that way but our `Url` is not prepared to manage `turn:trn530n002.karere.mega.nz:3478?transport=udp` as a url. `turn:` is not a valid part for a url. But I can removed that part and keep the rest 
I'll modify the PR

src/rtcModule/webrtc.cpp
alber2510: I'll add it

src/rtcModule/webrtc.cpp
alber2510: I use a `setIceServer` instead of `addIceServer` to ensure webrtc use the url returned by the API. If I use `add`, webrtc can use as first option one of the oldest

src/rtcModule/webrtc.cpp
alber2510: Yes, I think it's possible, I will modify

src/rtcModule/webrtc.cpp
alber2510: It isn't related to reduce latency but I paid attention during development. If you think is better remove to other PR I remove the line.
I added that condition because when there are invalid Turn server the session terminate with that termCode, then it's necessary to increase mIceFails and use static turn server the next time session start up

src/rtcModule/webrtc.cpp

src/rtcModule/webrtc.cpp

alber2510:
trigger compilation

src/rtcModule/webrtc.cpp
alber2510: I follow this approach to keep to avoid look in cache the correspondence between URL and `Shard number`

src/rtcModule/webrtc.cpp
```diff
@@ -940,82 +971,58 @@ void RtcModule::abortCallRetry(Id chatid)
...
+                             ipsv6.size() ? ipsv6[0].c_str() : "");
+                mKarereClient.mDnsCache.setIp(shard, ipsv4, ipsv6);
+            }
 
-                StaticProvider iceServer;
-                string jsonCacheTurnServer = getCachedTurnServers();
-                iceServer.setServer(jsonCacheTurnServer.c_str());
-                setIceServers(iceServer);
-            });
+            updateTurnServer();
```
sergiohs84: Do we actually update anything in the TURN servers passed to WebRTC when the `wsResolveDNS()` fails?

alber2510: Fixed it

alber2510:
trigger compilation

src/rtcModule/webrtc.cpp
```diff
@@ -87,15 +89,60 @@ RtcModule::RtcModule(karere::Client& client, IGlobalHandler& handler,
...
             return;
-        setIceServers(mIceServerProvider);
+
+        int shard = TURNSERVER_SHARD;
+        for (const std::shared_ptr<TurnServerInfo>& serverInfo : mIceServerProvider)
+        {
+            std::string fullUrl = serverInfo->url;
+            if (fullUrl.size())
+            {
+                size_t posInitialColon = fullUrl.find(":") + 1;
```
sergiohs84, line 117: An example of the `url` here, to understand the format, will help :)

src/rtcModule/webrtc.cpp
```diff
@@ -152,6 +199,75 @@ void RtcModule::removeCallRetry(karere::Id chatid, bool retry)
...
+string RtcModule::getCachedTurnServers()
+{
+    int i = 0;
+    vector<string> urls;
+    while (mKarereClient.mDnsCache.isValidUrl(TURNSERVER_SHARD - i) && i < MAX_TURN_SERVERS)
+    {
+        string ipv4;
+        string ipv6;
+        karere::Url turnServerUrl = mKarereClient.mDnsCache.getUrl(TURNSERVER_SHARD - i);
+        if (mKarereClient.mDnsCache.getIp(TURNSERVER_SHARD, ipv4, ipv6))
```
sergiohs84, line 211:
```suggestion
        if (mKarereClient.mDnsCache.getIp(TURNSERVER_SHARD - i, ipv4, ipv6))
```

src/rtcModule/webrtcPrivate.h
```diff
@@ -486,6 +494,39 @@ class RtMessageComposer: public chatd::Command
...
+    {
+        if (!instance)
+        {
+            instance = new DnsRequest();
+        }
+
+        return instance;
+    }
+
+    virtual ~DnsRequest() {}
```
sergiohs84, line 515: Shouldn't `instance` become `null`?? (the global variable).

src/rtcModule/webrtc.cpp
alber2510: Fixed

src/rtcModule/webrtcPrivate.h
alber2510: It's done at file webrtc.cpp line  50

sergiohs84: That's the initialization of the value. I mean at dtor. Anyway, there's no need of a Singleton here, simply add the `DnsRequest` as a member of `RtcModule`.

PR #827, opened 2020-03-27
#15019.Import new messages from a different cache

sergiohs84:


sergiohs84:
trigger compilation

src/chatd.cpp
```diff
@@ -4039,148 +4064,153 @@ void Chat::onMsgUpdated(Message* cipherMsg)
     })
     .then([this, updateTs, richLinkRemoved](Message* msg)
     {
-        assert(!msg->isPendingToDecrypt()); //either decrypted or error
```
sergiohs84: All this `.then()` block has been moved, without changes, to `onMsgUpdatedAfterDecrypt()`, so it can be called directly for imported messages.

sergiohs84:
Wrong diff. Recreating this PR...

src/chatClient.cpp
```diff
@@ -396,13 +396,28 @@ int Client::importMessages(const char *externalDbPath)
...
         std::string query;
         if (!chat.empty())
         {
             newestAppIdx = chat.highnum();
             newestAppMsg = &chat.at(newestAppIdx);
             newestAppMsgid = firstMsgidToImport = newestAppMsg->id();
+            uint32_t tsMissingUpdates = newestAppMsg->ts - CHATD_MAX_EDIT_AGE;
+
+            chatd::Idx messageIdx  = chat.highnum() - 1;
+            while (messageIdx > chat.lownum())
```
sergiohs84, line 410: It will only consider messages loaded in RAM, which is very likely enough for most cases, but it could happen that some messages in DB (not loaded yet) have been updated too.

Also, this block could be avoided in case the app's DB is ahead of NSE's DB, or in case there is a truncate.

src/chatClient.cpp
```diff
@@ -514,6 +535,36 @@ int Client::importMessages(const char *externalDbPath)
...
+        for (auto it = messagesUpdated.begin(); it != messagesUpdated.end(); it++)
+        {
+            query = "select userid, ts, type, data, idx, keyid, backrefid, updated, is_encrypted from history"
+                                " where chatid = ?1 and msgid = ?2 and updated != ?3";
+
+            karere::Id msgid = it->first;
+            uint16_t updated = it->second;
+            SqliteStmt stmtMsgUpdated(db, query.c_str());
+            stmtMsgUpdated << chatroom->chatid() << msgid << updated;
+            while (stmtMsgUpdated.step())
```
sergiohs84, line 548: This `while` is actually an `if`, IMHO.

sergiohs84:
Finally, I suggest to have a `countAdded` and `countUpdated` and return both of them, but in the log, indicate them separately.

src/chatClient.cpp

src/chatClient.cpp

src/chatClient.cpp
```diff
@@ -531,7 +526,7 @@ int Client::importMessages(const char *externalDbPath)
             }
 
             chat.msgImport(move(msg), isUpdate);
-            count++;
+            countAdded++;
```
sergiohs84, line 529:
```suggestion
            (isUpdate) ? countUpdated++ : countAdded;
```

src/megachatapi.h
```diff
@@ -2230,15 +2230,21 @@ class MegaChatApi
      * of the app's cache, but may include new messages that wants to be imported into the app's
      * cache in one shot. In case the history has been truncated, this method applies truncation.
      *
+     * The associated request type with this request is MegaChatRequest::TYPE_IMPORT_MESSAGES
+     * Valid data in the MegaChatRequest object received on callbacks:
+     * - MegaChatRequest::getText - Returns the cache path
```
sergiohs84, line 2235: We could add the count of imported messages to the request, so the app still can check. I'll do it.

src/megachatapi_impl.cpp
```diff
@@ -1768,6 +1768,27 @@ void MegaChatApiImpl::sendPendingRequests()
...
+            int count = mClient->importMessages(request->getText());
+            if (count < 0)
+            {
+                API_LOG_WARNING("Error importing Messages: %d", count);
+                errorCode = MegaChatError::ERROR_UNKNOWN;
+                break;
+            }
+
+            API_LOG_DEBUG("%d Messages has been imported", count);
+            errorCode = MegaChatError::ERROR_OK;
```
sergiohs84, line 1789: Caution! this request will never finish, since the `ERROR_OK == 0`, so no `fireOnRequestFinish()` would be called. I'll fix it ;)

src/chatClient.cpp

src/megachatapi_impl.cpp

src/megachatapi.h
alber2510: Fixed

PR #839, opened 2020-04-15
WIP: Update static TURN server

sergiohs84:
TBD if this change is actually needed, or sysadmins will get the server back.


sergiohs84:
Sysadmins have added a CNAME from `trn530n001` to `trn530n002`, so we don't need to change the static servers.

PR #842, opened 2020-04-28
#12792. Remove legacy encryption mode based on RSA

jgandres:
**Important:** don't merge this PR until #721 it's merged. Once #721 has been merged update this PR with develop.


src/strongvelope/strongvelope.cpp
```diff
@@ -192,12 +187,7 @@ bool ParsedMessage::verifySignature(const StaticBuffer& pubKey, const SendKey& s
     assert(pubKey.dataSize() == 32);
     if (protocolVersion < 2)
     {
-        //legacy
-        Buffer messageStr(SVCRYPTO_SIG.size()+signedContent.dataSize());
-        messageStr.append(SVCRYPTO_SIG.c_str(), SVCRYPTO_SIG.size())
-        .append(signedContent);
-        return (crypto_sign_verify_detached(signature.ubuf(), messageStr.ubuf(),
-                messageStr.dataSize(), pubKey.ubuf()) == 0);
+        return false;
```
alber2510, line 190: Maybe an assert is better instead of a `if (protocolVersion < 2) and a return false`

jgandres: I think that we could try to verify a signature of a very old message and in this case we should only return false

src/chatClient.cpp
```diff
@@ -1697,23 +1697,7 @@ promise::Promise<void> Client::loadOwnKeysFromApi()
...
-        auto privrsa = result->getPrivateKey();
-        if (!privrsa)
-            return ::promise::Error("No private RSA key in getUserData API response");
-        mMyPrivRsaLen = base64urldecode(privrsa, strlen(privrsa), mMyPrivRsa, sizeof(mMyPrivRsa));
-        // write to db
-        db.query("insert or replace into vars(name, value) values('pr_cu25519', ?)", StaticBuffer(mMyPrivCu25519, sizeof(mMyPrivCu25519)));
-        db.query("insert or replace into vars(name, value) values('pr_ed25519', ?)", StaticBuffer(mMyPrivEd25519, sizeof(mMyPrivEd25519)));
-        db.query("insert or replace into vars(name, value) values('pub_rsa', ?)", StaticBuffer(mMyPubRsa, mMyPubRsaLen));
-        db.query("insert or replace into vars(name, value) values('pr_rsa', ?)", StaticBuffer(mMyPrivRsa, mMyPrivRsaLen));
+        api.call(&mega::MegaApi::getUserData);
```
alber2510, line 1700:
```suggestion
```

src/karereCommon.cpp
```diff
@@ -14,7 +14,7 @@ namespace rtcModule {void globalCleanup(); }
 
 namespace karere
 {
-const char* gDbSchemaVersionSuffix = "9";
+const char* gDbSchemaVersionSuffix = "11";
```
sergiohs84, line 17: What about the version 10?? is it used by another open PR?

src/strongvelope/strongvelope.h
```diff
@@ -94,11 +94,6 @@ enum
 
 /** Contains a 64 bit user handle of a user that was excluded from chat */
     TLV_TYPE_EXC_PARTICIPANT    = 0x09,
-
-/** In case we encrypt keys using RSA, we can't reverse-decrypt them like with
- * EC, so we need to encrypt our own keys yo ourself so we can read our own
- * history messages. (legacy only) */
-    TLV_TYPE_OWN_KEY            = 0x0a,
```
sergiohs84, line 101: When you have enums that become obsolete/deprecated, it's better to leave traces in the code. In example, comment the line and add a comment-in-line: _deprecated (used for legacy encryption)_

sergiohs84: Otherwise, someone might consider as a typo and re-adjust the values, assigning `TLV_TYPE_INVITOR = 0x0a`.

src/strongvelope/strongvelope.h
```diff
@@ -109,8 +104,6 @@ enum
 /** Message types used for the chat message transport. */
 enum: unsigned char
 {
-    ///Legacy message containing a sender key (initial, key rotation, key re-send).
-    SVCRYPTO_MSGTYPE_KEYED                     = 0x00,
```
sergiohs84, line 113: Same here

src/strongvelope/tlvstore.h
```diff
@@ -50,7 +50,7 @@ class TlvParser
     size_t mOffset;
     bool mLegacyMode;
 public:
-    TlvParser(const StaticBuffer& source, size_t offset, bool legacyMode)
+    TlvParser(const StaticBuffer& source, size_t offset, bool legacyMode = false)
```
sergiohs84, line 53: Why don't we remove the param totally?

src/userAttrCache.h
```diff
@@ -22,13 +22,10 @@ namespace karere
  */
 enum {
     /** First attribute that is not directly supported by the SDK via getUserAttribute()*/
-    USER_ATTR_VIRTUAL_FIRST = 64,
-
-    /** RSA Public key of user */
-    USER_ATTR_RSA_PUBKEY = USER_ATTR_VIRTUAL_FIRST,
+    USER_ATTR_VIRTUAL_FIRST = 65,
```
sergiohs84, line 25: `64` was probably not a random value. Please, keep it and do not remove the value in the enum of callbacks. Simply mark the `USER_ATTR_RSA_PUBKEY` as deprecated, with a comment, so 64 simply becomes unused.

src/karereCommon.cpp
sergiohs84: IC, in the https://github.com/meganz/MEGAchat/pull/721

src/dbSchema.sql
```diff
@@ -32,7 +32,7 @@ CREATE TABLE history(idx int not null, chatid int64 not null, msgid int64 not nu
     userid int64, keyid int not null, type tinyint, updated smallint, ts int,
     is_encrypted tinyint, data blob, backrefid int64 not null, UNIQUE(chatid,msgid), UNIQUE(chatid,idx));
 
-CREATE TABLE sendkeys(chatid int64 not null, userid int64 not null, keyid int64 not null, key blob not null,
+CREATE TABLE sendkeys(chatid int64 not null, userid int64 not null, keyid int32 not null, key blob not null,
```
sergiohs84, line 35: Not sure if the migration of the DB to a newer version is really worthy

src/chatClient.cpp
```diff
@@ -320,6 +320,35 @@ bool Client::openDb(const std::string& sid)
                 ok = true;
                 KR_LOG_WARNING("Database version has been updated to %s", gDbSchemaVersionSuffix);
             }
+            else if (cachedVersionSuffix == "10" && (strcmp(gDbSchemaVersionSuffix, "11") == 0))
+            {
+                // Remove USER_ATTR_RSA_PUBKEY attr from cache
+                db.query("delete from userattrs where type = 64");
+
+                // Create temporary table and copy sendkeys content
+                db.query("CREATE TABLE tempkeys(chatid int64 not null, userid int64 not null, keyid int32 not null, key blob not null,ts int not null, UNIQUE(chatid, userid, keyid));");
```
sergiohs84, line 329: This way of moving `int64` values into records with `int32` values sounds risky. If someone had legacy key ids (which I hesitate), it could create issues. But given that none should have those keys at these days and it's already implemented... let's keep it. :)

src/chatClient.cpp
```diff
@@ -320,6 +320,35 @@ bool Client::openDb(const std::string& sid)
...
+                db.query("INSERT INTO tempkeys(chatid, userid, keyid, key, ts) SELECT chatid, userid, keyid, key, ts FROM sendkeys");
+
+                // Close and re-open db again to avoid SQLITE_LOCKED
```
sergiohs84, line 332: I don't get this one. Why is it locked?

src/chatClient.cpp
```diff
@@ -320,6 +320,35 @@ bool Client::openDb(const std::string& sid)
...
+                db.close();
+                if (!db.open(path.c_str(), false))
+                {
```
sergiohs84, line 335: Better to enclose the queries in the bloc if open is successful, and fallback to the standard error's handling routine in case it fails.

src/userAttrCache.h
jgandres: This change it's related with this one:  db.query("delete from userattrs where type = 64");
Previously USER_ATTR_RSA_PUBKEY = USER_ATTR_VIRTUAL_FIRST = 64
We have removed all userattrs where type is 64 from cache to avoid to retrieve them.

Now  USER_ATTR_EMAIL = USER_ATTR_VIRTUAL_FIRST = 65

src/dbSchema.sql
jgandres: Please check ticket where it's the link to full conversation about this topic.

jgandres: Please check ticket to see the conversation link about this topic.

src/chatClient.cpp
jgandres: Ok

src/chatClient.cpp
jgandres: I've followed the same approach as here: https://github.com/meganz/MEGAchat/blob/feature/legacy-encryption-code-cleanup/src/chatClient.cpp#L169

src/chatClient.cpp
jgandres: https://ti-qa-archive.github.io/question/119597/sqlite-error-database-table-is-locked.html

The database file will lock when a write action is being  performed such as a insert, delete or update. You'll need to ensure that  you close and/or end the transaction of any insert/delete/update  queries to prevent a locking run-time error. It's also worth noting that  currently a try-catch around the query execution will not catch any SQL  error notice.

src/chatClient.cpp
sergiohs84: Yes, but you already opened the DB started to migrate the squema. Also, the log message is the same than the previous one (always try to make log messages unique in the code, so it's easier for debugging).

src/strongvelope/tlvstore.h
```diff
@@ -48,20 +48,19 @@ class TlvParser
...
-    TlvParser(const StaticBuffer& source, size_t offset, bool legacyMode = false)
-    :mSource(source), mOffset(offset), mLegacyMode(legacyMode){}
+    TlvParser(const StaticBuffer& source, size_t offset)
+    :mSource(source), mOffset(offset){}
     bool getRecord(TlvRecord& record)
     {
        if (mOffset == Buffer::kNotFound)
             return false;
-        size_t typeLen = mLegacyMode ? 2 : 1;
+        size_t typeLen = 1;
```
sergiohs84, line 58: Now `typeLen` is always `1`, since the `sizeof(uint8_t)` is 1 byte. Shouldn't we adjust the code accordingly, instead of allocating a variable just for the shake of assigning it to a constant?

src/userAttrCache.h
sergiohs84: Better to restore the 64, even if not used

src/chatClient.cpp
sergiohs84: No other thread should be reading from this DB. We are initializing it, even before anyone else can write or read from this DB, so it shouldn't be locked. Please, retest without the `close` + `open` and ensuring no other DB connections are opened.

jgandres:
trigger compilation

PR #844, opened 2020-04-29
#15448. Take into account missed call management messages as unread count message

alber2510:


src/chatdDb.h
```diff
@@ -284,7 +284,36 @@ class ChatdSqliteDb: public chatd::DbInterface
         if (idx != CHATD_IDX_INVALID)
             stmt << idx;
         stmt.stepMustHaveData("get peer msg count");
-        return stmt.intCol(0);
+        int32_t unReadCount = stmt.intCol(0);
+
+        sql = "select data from history where (chatid = ?1)"
+                "and (userid != ?2)"
+                "and not (updated != 0 and length(data) = 0)"
+                "and (is_encrypted = ?3)"
```
sergiohs84, line 292: IIRC, management message are not encrypted. If so, we could skip this condition.

src/chatdDb.h
```diff
@@ -284,7 +284,36 @@ class ChatdSqliteDb: public chatd::DbInterface
         if (idx != CHATD_IDX_INVALID)
             stmt << idx;
         stmt.stepMustHaveData("get peer msg count");
-        return stmt.intCol(0);
+        int32_t unReadCount = stmt.intCol(0);
+
+        sql = "select data from history where (chatid = ?1)"
+                "and (userid != ?2)"
+                "and not (updated != 0 and length(data) = 0)"
```
sergiohs84, line 291: Are `CallEnd` messages editable or even deletable? I don't think so, so we can skip those conditions too.

src/chatdDb.h
```diff
@@ -284,7 +284,36 @@ class ChatdSqliteDb: public chatd::DbInterface
         if (idx != CHATD_IDX_INVALID)
             stmt << idx;
         stmt.stepMustHaveData("get peer msg count");
-        return stmt.intCol(0);
+        int32_t unReadCount = stmt.intCol(0);
+
+        sql = "select data from history where (chatid = ?1)"
+                "and (userid != ?2)"
```
sergiohs84, line 290: `userid` for management messages is the `COMMANDER`, isn't it? If so, we can actually search by `userid == COMMANDER`, but with the `type` and `chatid` should be enough (plus the `idx`)

src/chatdMsg.h
```diff
@@ -749,6 +758,16 @@ class Message: public Buffer
     // ones in ChatdSqliteDb::getUnreadMsgCountAfterIdx()
     bool isValidUnread(karere::Id myHandle) const
     {
+        bool validCallEndMessage = false;
+        if (type == kMsgCallEnd)
```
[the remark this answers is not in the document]
alber2510: I'll add a commit with the change and review if it's ok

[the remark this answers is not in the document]
alber2510: Fixed

src/chatdDb.h
sergiohs84: Alright, then we do have to keep the condition for the `userid != me` :) 
Thanks for checking

src/chatdMsg.h
```diff
@@ -789,6 +779,35 @@ class Message: public Buffer
         return (type == kMsgContainsMeta && dataSize() > 3) ? std::string(buf()+3, dataSize() - 3) : "";
     }
 
+    bool isMissingCall(karere::Id myHandle) const
+    {
+        if (type != kMsgCallEnd)
```
sergiohs84, line 784:
```suggestion
        if (type != kMsgCallEnd || isOwnMessage(myHandle))
```

src/chatdMsg.h
```diff
@@ -789,6 +779,35 @@ class Message: public Buffer
...
+        {
+            return false;
+        }
+
+        uint8_t termCode = Message::extractTermCodeEndCall(*this);
+
+        return !isOwnMessage(myHandle) && (termCode == CallDataReason::kNoAnswer || termCode == CallDataReason::kCancelled);
```
sergiohs84, line 791:
```suggestion
        return (termCode == CallDataReason::kNoAnswer || termCode == CallDataReason::kCancelled);
```

src/chatdMsg.h
```diff
@@ -789,6 +779,35 @@ class Message: public Buffer
...
+
+    static uint8_t extractTermCodeEndCall(const Buffer& buffer)
+    {
+        unsigned int lenCallid = sizeof (karere::Id);
+        unsigned int lenDuration = sizeof (uint32_t);
+        unsigned int lenTermCode = sizeof (uint8_t);
+
+        if (buffer.size() < (lenCallid + lenDuration + lenTermCode))
+        {
+            return -1;
```
sergiohs84, line 802: This method returns unsigned int
```suggestion
            return 0xFF;
```

PR #850, opened 2020-05-06
#12074. When a message of a chat is updated without an internet connection, the last message is not updated

jgandres:


src/chatd.cpp
```diff
@@ -5182,11 +5197,20 @@ bool Chat::findLastTextMsg()
             auto& msg = *it->msg;
             if (msg.isValidLastMessage())
             {
+                if (deletedMsg.find(msg.id()) != deletedMsg.end())
+                {
+                    // Skip lastTextMsg candidate because there's another item to remove this msg in the queue
```
sergiohs84, line 5202: When a message wants to be deleted, doesn't the code update the content of all related items?? In example, the queue has M1, M2, M1.1. When the user attempts to delete M2 -> M2.1 (empty content), the content of M2 is also deleted, if still in the queue. Isn't it?
If the above is correct (please, check it), the `if (msg.isValidLastMessage()) will fail, as it will the `else if ()` below. In consequence, all these changes will never run :thinking-face:

jgandres: The above is correct so this commit is invalid.

PR #851, opened 2020-05-06
#15581. Assert (msg->isSending()) failed in Chat::msgRemoveFromSending

jgandres:


src/chatdDb.h
```diff
@@ -131,8 +131,8 @@ class ChatdSqliteDb: public chatd::DbInterface
     virtual int updateSendingItemsMsgidAndOpcode(karere::Id msgxid, karere::Id msgid)
     {
         mDb.query(
-            "update sending set opcode=?, msgid=? where chatid=? and opcode=? and msgid=?",
-            chatd::OP_MSGUPD, msgid, mChat.chatId(), chatd::OP_MSGUPDX, msgxid);
+            "update sending set opcode=?, msgid=? where chatid=? and (opcode=? or opcode=?) and msgid=?",
```
sergiohs84, line 134: I think this is wrong. Since we are updating the `msgid`, it means we already received the `NEWMSGID`, so no `NEWMSG` should be in the sending table anymore. It should have been removed from a previous call to `deleteSendingItem()` (from `msgRemoveFromSending()`).

PR #854, opened 2020-05-08
#15563. Import SEEN from external cache

jgandres:


src/chatClient.cpp
```diff
@@ -393,6 +393,42 @@ int Client::importMessages(const char *externalDbPath)
...
+        stmtLastSeen << chatid;
+        stmtLastSeen.stepMustHaveData();
+        lastSeenId = stmtLastSeen.uint64Col(0);
+        chat.setLastSeenId(lastSeenId);
+
+        SqliteStmt stmtLastSeenIdx(dbExternal, "select idx from history where chatid = ? and msgid = ?");
+        stmtLastSeenIdx << chatid << lastSeenId;
+        if (stmtLastSeenIdx.step())
+        {
+            chat.setLastSeenIdx(stmtLastSeenIdx.intCol(0));
```
sergiohs84, line 408: This is risky. The `idx` given by the NSE to an specific `msgid` doesn't necessarily match the `idx` given by the app to the same message.

src/chatClient.cpp
```diff
@@ -393,6 +393,42 @@ int Client::importMessages(const char *externalDbPath)
...
+        }
+        chatroom->chat().handleLastImportedSeen(lastSeenId);
+
+        // Import last received from external db
+        Id lastRecvId;
+        SqliteStmt stmtRecv(dbExternal, "select last_recv from chats where chatid=?");
```
sergiohs84, line 414: No clients are using/updating the `RECEIVED`, so probably we can simply skip this part.
I'll check with the team if there are plans to restore the feature in a future. In that case, better to keep it.

src/chatClient.cpp
```diff
@@ -393,6 +393,42 @@ int Client::importMessages(const char *externalDbPath)
         chatd::Chat &chat = chatroom->chat();
         karere::Id chatid = chatroom->chatid();
 
+        // Import last seen from external db
+        Id lastSeenId;
+        SqliteStmt stmtLastSeen(dbExternal, "select last_seen from chats where chatid=?");
+        stmtLastSeen << chatid;
+        stmtLastSeen.stepMustHaveData();
```
sergiohs84, line 400: What if the chatroom has no history?? In that case, I guess there's no  `SEEN` for the given `chatid`

src/chatClient.cpp
```diff
@@ -393,6 +393,42 @@ int Client::importMessages(const char *externalDbPath)
...
+        chat.setLastRecvId(lastRecvId);
+
+        SqliteStmt stmtRecvIdx(dbExternal, "select idx from history where chatid = ? and msgid = ?");
+        stmtRecvIdx << chatid << lastRecvId;
+        if (stmtRecvIdx.step())
+        {
+            chat.setLastRecvIdx(stmtRecvIdx.intCol(0));
+        }
+
+        if (chatroom->chat().lastSeenId() != chatroom->chat().lastReceivedId())
```
sergiohs84, line 427: What's the goal of comparing the last message seen by own user, and the last message received by other user?

src/chatClient.cpp
```diff
@@ -393,6 +393,42 @@ int Client::importMessages(const char *externalDbPath)
...
+        {
+            chatroom->chat().handleLastImportedSeen(lastRecvId);
+        }
+
```
sergiohs84, line 431: Would it work to call `Chat::onLastSeen()`?? as if the `SEEN` had been received from chatd. It should do all the job already.

sergiohs84:
Closing this approach in pro of https://github.com/meganz/MEGAchat/pull/858

PR #866, opened 2020-05-15
#15510. Improve addition / deletion of peers to presenced

jgandres:


jgandres:
trigger compilation

jgandres:
trigger compilation

src/presenced.cpp
```diff
@@ -606,7 +606,6 @@ void Client::onChatsUpdate(::mega::MegaApi *api, ::mega::MegaTextChatList *rooms
                 {
                     uint64_t userid = peerList->getPeerHandle(j);
                     mChatMembers[chatid].insert(userid);
-                    addPeer(userid);
```
sergiohs84, line 609: Now that we don't need to track the peers of each chatroom, I think all the processing happening in `onChatsUpdate()` becomes obsolete and useless. That's why you removed the only purpose of the implementation of this callback: `addPeer()` and `removePeer()`.
I'll adjust it

PR #867, opened 2020-05-15
#15729.  Avoid to notify about messages being decrypted

sergiohs84:
When a new message is received, but it is already SEEN, the `onChatNotification()` should not be called in case the message is still being decrypted. Once decrypted, a callback, if suitable, will be
triggered.

### DIFF
--- a/src/megachatapi_impl.cpp
+++ b/src/megachatapi_impl.cpp
@@ -4551,8 +4551,11 @@ void MegaChatApiImpl::onInitStateChange(int newState)
 
 void MegaChatApiImpl::onChatNotification(karere::Id chatid, const Message &msg, Message::Status status, Idx idx)
 {
-    if (megaApi->isChatNotifiable(chatid))
-     {
+    if (megaApi->isChatNotifiable(chatid)
+            && msg.userid != chatApi->getMyUserHandle()
+            && status == chatd::Message::kSeen  // received message from a peer changed to seen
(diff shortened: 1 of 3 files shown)


src/megachatapi_impl.cpp
```diff
@@ -4551,8 +4551,11 @@ void MegaChatApiImpl::onInitStateChange(int newState)
 
 void MegaChatApiImpl::onChatNotification(karere::Id chatid, const Message &msg, Message::Status status, Idx idx)
 {
-    if (megaApi->isChatNotifiable(chatid))
-     {
+    if (megaApi->isChatNotifiable(chatid)
```
alber2510, line 4554: If we add this check here instead of `ChatRoom::onMessageStatusChange`, I think we miss the other usages of `onChatNotification`:
 - `ChatRoom::onRecvNewMessage(chatd::Idx idx, chatd::Message& msg, chatd::Message::Status status)`
 - `void ChatRoom::onMessageEdited(const chatd::Message& msg, chatd::Idx idx)`
In both methods, we check `status == chatd::Message::kNotSeen` to call `parent.mKarereClient.app.onChatNotification(mChatid, msg, status, idx)` then for this other cases we miss the call to `fireOnChatNotification(chatid, message);`

sergiohs84: Good catch!! Thanks, Alber. Please, see my next commit: `15dd68bf4ddc2d178790b3e4af827bdf886f6bc6`

PR #871, opened 2020-05-20
 #10235. Retention time for history

jgandres:


jgandres:
trigger compilation

jgandres:
trigger compilation

src/megachatapi_impl.cpp
```diff
@@ -1792,6 +1792,46 @@ void MegaChatApiImpl::sendPendingRequests()
...
+            int period = request->getParamType();
+            bool inSeconds = request->getFlag();
+
+            if (chatid == MEGACHAT_INVALID_HANDLE)
+            {
+                errorCode = MegaChatError::ERROR_ARGS;
+                break;
+            }
+
+            ChatRoom *chatroom = findChatRoom(chatid);
```
alber2510, line 1807: I think we have to check if period is greater or equal to zero with
I we don't check that, we have to prepare to handle retention time negative

jgandres: To disable retention time you can provide zero or negative number.
I forgot to update the documentation, sorry about that. I've changed at
1c55e12

alber2510: We are handling the value received from chatd as `unsigned int`, if we set a negative value, chatd returns 0 or negative?

jgandres: Chatd returns 0

alber2510: Then, it's perfect

src/chatd.cpp
```diff
@@ -4375,6 +4500,23 @@ void Chat::deleteMessagesBefore(Idx idx)
...
 
+void Chat::truncateByRetentionTime(Idx idx)
+{
+    if (idx >= mForwardStart)
+    {
+        mBackwardList.clear();
+        auto delCount = idx - mForwardStart;
+        auto end = mForwardList.begin() + delCount;
+        mForwardList.erase(mForwardList.begin(), end);
+        mForwardList.erase(mForwardList.begin());
```
sergiohs84, line 4511: Isn't it the same than... ??
```suggestion
        mForwardList.erase(mForwardList.begin(), end + 1);
```

src/chatd.cpp
```diff
@@ -4454,7 +4596,9 @@ Idx Chat::msgIncoming(bool isNew, Message* message, bool isLocal)
             if (mHasMoreHistoryInDb)
             { //we have db history that is not loaded, so we determine the index
               //by the db, and don't add the message to RAM
-                idx = mDbInterface->getOldestIdx()-1;
+                chatd::Idx oldestIdx = mDbInterface->getOldestIdx();
+                assert(oldestIdx != CHATD_IDX_INVALID);
+                idx = oldestIdx - 1;
```
sergiohs84, line 4601:
```suggestion
                assert(mDbInterface->getOldestIdx() != CHATD_IDX_INVALID);
                idx = mDbInterface->getOldestIdx() - 1;
```

src/chatd.cpp
```diff
@@ -4735,6 +4879,10 @@ void Chat::msgIncomingAfterDecrypt(bool isNew, bool isLocal, Message& msg, Idx i
             mAttachmentNodes->addMessage(msg, isNew, false);
         }
         CALL_DB(addMsgToHistory, msg, idx);
+        if (!isNew || mOldestIdxInDb == CHATD_IDX_INVALID)
+        {
+            mOldestIdxInDb = mDbInterface->getOldestIdx();
```
sergiohs84, line 4884: Again, I think here you should be able to find the oldest idx without the need of the DB query, right?

src/chatd.h
```diff
@@ -124,6 +124,17 @@ class Listener
...
+    /**
+     * @brief Chat history have been truncated as result of reception of OP_RETENTION command.
+     * All messages that exceed retention time must be cleared, so we need to inform app
+     * to free all resources associated to these messages.
+     *
+     * @param msg Most recent message affected by retention time
+     * @param idx Index of the message
+     * @param status Status of the message
+     */
+    virtual void onRetentionHistoryTruncated(const Message &msg, const Idx &idx, const Message::Status &status) {}
```
sergiohs84, line 136: It looks to me more natural:
```suggestion
    virtual void onHistoryTruncatedByRetentionTime(const Message &msg, const Idx &idx, const Message::Status &status) {}
```

src/chatd.h
```diff
@@ -840,6 +855,8 @@ class Chat: public karere::DeleteTrackable
     bool mTruncateAttachment = false;
     /** Indicates the reaction sequence number for this chatroom */
     karere::Id mReactionSn = karere::Id::inval();
+    /** Indicates the retention time for this chat room, after which the previous messages are automatically deleted **/
```
sergiohs84, line 858:
```suggestion
    /** Indicates the retention time for this chat room, after which the previous messages are automatically deleted */
```

src/chatd.cpp
```diff
@@ -4329,7 +4360,100 @@ void Chat::handleTruncate(const Message& msg, Idx idx)
...
+    // Discard current key, if any
+    CALL_CRYPTO(resetSendKey);
+
+    // Clean affected messages in db and RAM
+    CHATID_LOG_DEBUG("Cleaning messages previous to %d seconds", retentionTime);
+    CALL_DB(retentionHistoryTruncate, idx);
+    truncateByRetentionTime(idx);
+
+    // Clean pending rich links
+    removePendingRichLinks(idx);
```
sergiohs84, line 4401: You may need to remove pending reactions, in the corresponding branch. I guess this branch will be merged first.

src/chatdDb.h
```diff
@@ -381,7 +381,15 @@ class ChatdSqliteDb: public chatd::DbInterface
         SqliteStmt stmt(mDb, "select min(idx) from history where chatid = ?");
         stmt << mChat.chatId();
         stmt.stepMustHaveData(__FUNCTION__);
-        return stmt.uint64Col(0);
+
+        if (sqlite3_column_type(stmt, 0) == SQLITE_NULL) // no db history
```
sergiohs84, line 385: I think this change may break the existing usage of this method to initialize the `idx` used to add new messages backwards in `msgIncoming()`. What is its motivation?

src/chatdDb.h
```diff
@@ -604,6 +612,32 @@ class ChatdSqliteDb: public chatd::DbInterface
...
+
+        if (sqlite3_column_type(stmt, 0) != SQLITE_NULL)
+        {
+            limitIdx = stmt.intCol(1);
+        }
+        else
+        {
+            limitIdx = CHATD_IDX_INVALID;
+        }
+    }
```
sergiohs84, line 632:
```suggestion
    chatd::Idx getIdxByRetentionTime(const time_t ts) override
    {
        // Find the most recent msg affected by retention time, if any
        SqliteStmt stmt(mDb, "select MAX(ts), MAX(idx) from history where chatid = ? and ts < ?");
        stmt << mChat.chatId() << ts;
        return (stmt.step()) ? stmt.intCol(1) : CHATD_IDX_INVALID;
    }
```

tests/sdk_test/sdk_test.cpp
```diff
@@ -32,6 +32,8 @@ int main(int argc, char **argv)
     // Tests that requires a groupchat (start with public chat, converted into private)
     EXECUTE_TEST(t.TEST_PublicChatManagement(0, 1), "TEST Publicchat management");
     EXECUTE_TEST(t.TEST_GroupChatManagement(0, 1), "TEST Groupchat management");
+    // TODO: uncomment this test once the retention history support is deployed into shards 0 and 1 (currently, it only works in shard 2)
```
sergiohs84, line 35: Is it already in production?

src/megachatapi.h
```diff
@@ -5749,6 +5793,15 @@ class MegaChatRoomListener
...
     virtual void onReactionUpdate(MegaChatApi* api, MegaChatHandle msgid, const char* reaction, int count);
+
+    /**
+     * @brief This function is called when we need to clear messages previous to retention time,
+     * all messages previous to received msg as parameter must be cleared.
+     *
+     * @param api MegaChatApi connected to the account
+     * @param msg Most recent message whose timestamp has exceeded retention time
+     */
+    virtual void onRetentionHistoryTruncated(MegaChatApi* /*api*/, MegaChatMessage* /*msg*/);
```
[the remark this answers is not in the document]
jgandres: I think this one it's ok, impl version has been marked as override.

tests/sdk_test/sdk_test.cpp
jgandres: First we need to be able to set retention time to a few seconds (only for testing purposes) if we want to use this test.

src/chatd.cpp
jgandres: I think so, I've noted to make the adjustment in the corresponding branch once this have been merged

src/megachatapi.h
```diff
@@ -3676,6 +3688,31 @@ class MegaChatApi
...
+     * - MegaChatRequest::getFlag - Returns true if period is specified in seconds, otherwise returns false.
+     *
+     * On the onRequestFinish error, the error code associated to the MegaChatError can be:
+     * - MegaChatError::ERROR_ARGS - If the chatid is invalid
+     * - MegaChatError::ERROR_NOENT - If there isn't any chat with the specified chatid.
+     * - MegaChatError::ERROR_ACCESS - If the logged in user doesn't have operator privileges
+     *
+     * @param chatid MegaChatHandle that identifies the chat room
+     * @param period retention timeframe after which all messages in the chat are automatically deleted
+     * @param inSeconds indicate API that period is specified in seconds. Only for testing
```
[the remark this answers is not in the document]
jgandres: Pending to check if period is under 1h

src/chatClient.cpp
```diff
@@ -2697,6 +2697,16 @@ promise::Promise<void> ChatRoom::archiveChat(bool archive)
     });
 }
 
+promise::Promise<void> ChatRoom::setChatRetentionTime(int period, bool inSeconds)
+{
+    auto wptr = getDelTracker();
+    return parent.mKarereClient.api.callIgnoreResult(&::mega::MegaApi::setChatRetentionTime, chatid(), period, inSeconds)
+    .then([wptr]()
+    {
+        wptr.throwIfDeleted();
```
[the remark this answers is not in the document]
jgandres: Any weak pointer is declared in MegaChatApi_Impl so I decided to maintain .then block, but If there's no inconveniences I'll change it.

src/chatd.cpp
```diff
@@ -4329,7 +4360,100 @@ void Chat::handleTruncate(const Message& msg, Idx idx)
...
+
+    // update oldest index in db
+    mOldestIdxInDb = mDbInterface->getOldestIdx();
+
+    if (mOldestIdxInDb == CHATD_IDX_INVALID) // If there's no messages in db
+    {
+        mLastIdxReceivedFromServer = CHATD_IDX_INVALID;
+        mLastIdReceivedFromServer = karere::Id::null();
```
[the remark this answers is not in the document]
jgandres: As I can see in develop lastIdReceivedFromServer It's only set but not used for anymore, and the only one getter it's not used. If we don't use this var anymore could we remove it?

src/chatdDb.h
jgandres: Before this change the only usage for this function was in msgIncoming as you said but

src/chatdDb.h
jgandres: This query always will have a step, so we need the following condition to differentiate if query returned a valid row or NULL:
(sqlite3_column_type(stmt, 0) != SQLITE_NULL)

src/chatd.cpp
jgandres: In case assert doesn't fail, query will be executed twice to retrieve the same value

src/chatd.cpp
```diff
@@ -4329,7 +4360,100 @@ void Chat::handleTruncate(const Message& msg, Idx idx)
...
+    // Clean affected messages in db and RAM
+    CHATID_LOG_DEBUG("Cleaning messages previous to %d seconds", retentionTime);
+    CALL_DB(retentionHistoryTruncate, idx);
+    truncateByRetentionTime(idx);
+
+    // Clean pending rich links
+    removePendingRichLinks(idx);
+
+    // update oldest index in db
+    mOldestIdxInDb = mDbInterface->getOldestIdx();
```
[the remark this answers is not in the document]
jgandres: But can we ensure that index + 1 exists in DB for all cases?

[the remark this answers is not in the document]
jgandres: Can we ensure that idx+1 exists in DB for all cases?

src/chatd.cpp
sergiohs84: The suggestion aimed to avoid the unnecessary allocation and copy into `oldestIdx`, but keeping the assert. It's true that it represents an extra DB query, but only in debug builds.
If you prefer to keep the allocation+assignment, I'm fine with it too :)

src/megachatapi_impl.cpp
```diff
@@ -7572,6 +7669,15 @@ MegaChatMessagePrivate::MegaChatMessagePrivate(const Message &msg, Message::Stat
             }
             break;
         }
+
+        case MegaChatMessage::TYPE_SET_RETENTION_TIME:
+        {
+          // Interpret retentionTime as int32_t to store it in an existing member.
+          // There's no overflow risk since retentionTime shouldn't reach such high values
+          memcpy(&priv, msg.buf(), sizeof(int32_t));
```
sergiohs84, line 7677: Would it be possible that `msg.buf()` has less than 4 bytes? Not sure if it would be possible, but if it were the case, that message will create issues by reading out of the boundaries of `msg`.

tests/sdk_test/sdk_test.cpp
jgandres: Yes

src/chatdDb.h
sergiohs84: IC. You use the `max()` and that's why the step success (queries ALWAYS have an step and it may result on a `SQLITE_ROW` or on a `SQLITE_DONE`). What you mean is that in this case, even when there's no `MAX()` values, it still returns `SQLITE_ROW`, but the values are `SQLITE_NULL`. 
Your code, however, checks for `MAX(ts)` not being null, and then it reads the `MAX(idx)`, so it may be 0, which will result on unexpected behavior, since that index corresponds to a different message.

sergiohs84:
```suggestion
    void getIdxByRetentionTime(const time_t ts, chatd::Idx &limitIdx) override
    {
        // Find the most recent msg affected by retention time if any
        SqliteStmt stmt(mDb, "select MAX(ts), MAX(idx) from history where chatid = ? and ts < ?");
        stmt << mChat.chatId() << ts;
        return (stmt.step() && sqlite3_column_type(stmt, 1) != SQLITE_NULL) ? stmt.intCol(1) : CHATD_IDX_INVALID;
    }
```

src/chatd.cpp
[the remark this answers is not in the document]
sergiohs84: It looks it might be handy for the management of the `RECEIVED` pointer, so the `mLastIdxReceivedFromServer`, since the `idx` should correspond to the `msgid`.
However, as you said, the feature is currently disabled and not working, so I think it's fine to remove it. In the future, if the `RECEIVED` is enabled (only for 1on1), we would see if the `msgid` is needed.

src/chatd.cpp
```diff
@@ -3840,6 +3866,10 @@ Idx Chat::msgConfirm(Id msgxid, Id msgid, uint32_t timestamp)
     }
 
     CALL_DB(addMsgToHistory, *msg, idx);
+    if (mOldestIdxInDb == CHATD_IDX_INVALID)
+    {
+        mOldestIdxInDb = mDbInterface->getOldestIdx();
```
sergiohs84, line 3859: At this point, the history was empty and now it has one message that the client sent and just got confirmed, right?
In that case, you don't need a DB query:
```suggestion
        mOldestIdxInDb = idx;
```

src/chatd.cpp
```diff
@@ -1837,6 +1845,7 @@ void Chat::disable(bool state)
 Idx Chat::getHistoryFromDb(unsigned count)
 {
     assert(mHasMoreHistoryInDb); //we are within the db range
+    mOldestIdxInDb = mDbInterface->getOldestIdx();
```
sergiohs84, line 1848: Why do we need to update the oldest index in DB every time we load more history from DB? Does it change from a call to another? The only possible change I can imagine it that we loaded more messages backwards in history, and the index should get updated upon reception of the message, IMHO.

src/chatd.cpp
```diff
@@ -4279,6 +4297,7 @@ void Chat::handleTruncate(const Message& msg, Idx idx)
     CHATID_LOG_DEBUG("Truncating chat history before msgid %s, idx %d, fwdStart %d", ID_CSTR(msg.id()), idx, mForwardStart);
     CALL_CRYPTO(resetSendKey);      // discard current key, if any
     CALL_DB(truncateHistory, msg);
+    mOldestIdxInDb = mDbInterface->getOldestIdx();
```
sergiohs84, line 4294: Isn't now the oldest index in DB the `idx`?? We've just truncated all previous history. Why should we do a DB query then?

src/chatd.cpp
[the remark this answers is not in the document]
sergiohs84: You can check what's the highest index in history, which should be `highnum()`, so:
```suggestion
    mOldestIdxInDb = (idx + 1 <= highnum()) ? idx + 1 : CHATD_IDX_INVALID;
```

src/chatd.cpp
```diff
@@ -4454,7 +4579,8 @@ Idx Chat::msgIncoming(bool isNew, Message* message, bool isLocal)
             if (mHasMoreHistoryInDb)
             { //we have db history that is not loaded, so we determine the index
               //by the db, and don't add the message to RAM
-                idx = mDbInterface->getOldestIdx()-1;
+                assert(mDbInterface->getOldestIdx() != CHATD_IDX_INVALID);
+                idx = mDbInterface->getOldestIdx() - 1;
```
sergiohs84, line 4575: Actually, now that we keep the oldest index updated in a variable, I wonder if we could simply do a
```suggestion
                idx = mOldestIdxInDb - 1;
```

src/megachatapi.h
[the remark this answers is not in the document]
sergiohs84: I would not limit the minimum period of time. Anything >= 0 should be accepted. If the API decides to restrict it, we will simply receive the error :)

src/chatdDb.h
jgandres: Changes undone

sergiohs84:
@jgandres could you check how mobile apps in production cope with the corresponding management message? Apart from that check, I think we can proceed with the merge.

jgandres:
> @jgandres could you check how mobile apps in production cope with the corresponding management message? Apart from that check, I think we can proceed with the merge.

For mobile apps in production msg will be handled as unknown type of management message and msg will be hide

PR #874, opened 2020-05-21
#15279. Retry call when recieve our end call in some cases

alber2510:
If we are in a call and due a some missing Incall, we recieve an EndCall. If we aren't in a reconnection, we launch a new reconnection

### DIFF
--- a/src/rtcModule/webrtc.cpp
+++ b/src/rtcModule/webrtc.cpp
@@ -2621,7 +2621,21 @@ void Call::onClientLeftCall(Id userid, uint32_t clientid)
         }
 
         SUB_LOG_DEBUG("ENDCALL received for ourselves, finishing the call");
-        destroy(TermCode::kErrNetSignalling, false);
+        auto wptr = weakHandle();
+        destroy(TermCode::kErrNetSignalling, false)
+        .then([wptr, this]
+        {
+            karere::Id chatid = chat().chatId();
+            auto it = mManager.mRetryCall.find(chatid);
+            if (it == mManager.mRetryCall.end())
+            {
+                karere::AvFlags flags = sentAv();
+                mManager.mRetryCall[chatid] = std::pair<karere::AvFlags, bool>(flags, true);
+                auto itHandler = mManager.mCallHandlers.find(chatid);
+                mManager.joinCall(chatid, flags, *itHandler->second, mId);
+                mManager.mRetryCall.erase(chatid);
+            }
+        });
         return;
     }
 
--- a/src/rtcModule/webrtc.cpp
+++ b/src/rtcModule/webrtc.cpp
@@ -319,24 +319,6 @@ bool RtcModule::selectVideoInDevice(const string &devname)
     return false;
 }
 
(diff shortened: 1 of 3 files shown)


src/rtcModule/webrtc.cpp
```diff
@@ -2621,7 +2621,21 @@ void Call::onClientLeftCall(Id userid, uint32_t clientid)
         }
 
         SUB_LOG_DEBUG("ENDCALL received for ourselves, finishing the call");
-        destroy(TermCode::kErrNetSignalling, false);
+        auto wptr = weakHandle();
+        destroy(TermCode::kErrNetSignalling, false)
+        .then([wptr, this]
```
sergiohs84, line 2626: Why do we want a `wptr` if it's not checked in the lambda?? :stuck_out_tongue:

src/rtcModule/webrtc.cpp
```diff
@@ -2621,7 +2621,21 @@ void Call::onClientLeftCall(Id userid, uint32_t clientid)
         }
 
         SUB_LOG_DEBUG("ENDCALL received for ourselves, finishing the call");
-        destroy(TermCode::kErrNetSignalling, false);
+        auto wptr = weakHandle();
+        destroy(TermCode::kErrNetSignalling, false)
```
sergiohs84, line 2625: There is a third param to indicate the reason of this Call being destroyed, which could be a great value while debugging. Please, add the reason of destruction here. Actually, that message is missing in many other calls, so you may want to improve it all over the code (about 20 calls).

src/rtcModule/webrtc.cpp
```diff
@@ -2621,7 +2621,21 @@ void Call::onClientLeftCall(Id userid, uint32_t clientid)
...
+        .then([wptr, this]
+        {
+            karere::Id chatid = chat().chatId();
+            auto it = mManager.mRetryCall.find(chatid);
+            if (it == mManager.mRetryCall.end())
+            {
+                karere::AvFlags flags = sentAv();
+                mManager.mRetryCall[chatid] = std::pair<karere::AvFlags, bool>(flags, true);
```
sergiohs84, line 2633: You already looked for the `chatid` and it's available, simply use it.
```suggestion
                it->second = std::pair<karere::AvFlags, bool>(flags, true);
```

src/rtcModule/webrtc.cpp
alber2510: Fixed

src/rtcModule/webrtc.cpp
alber2510: Ok, I'll add a description

src/rtcModule/webrtc.cpp
alber2510: Yes, I looked for it but I didn't find it, then I added it.
We retry the call if there isn't other retry on fly

sergiohs84: Ouch :facepalm: You're totally right, I misread the test.

PR #877, opened 2020-05-25
#15795. When I want to obtain a participant's full name, I get the nickname

jgandres:


src/chatClient.cpp
```diff
@@ -4279,19 +4292,29 @@ void Client::updateAliases(Buffer *data)
...
             std::string title = getUserAlias(userid);
             if (title.empty())
             {
-                title = !contact->getContactName().empty()
-                    ? contact->getContactName()
-                    : contact->email();
+                if (!contact->getContactName().empty())
+                {
+                    title = contact->getContactName();
+                    needEncoding = false;
```
sergiohs84, line 4305: I think that here we will set the title to `<firstname><lastname>`, without a space separator, which would be a bug.
I feel the best bet is to abandon the binary layout for the title, it should be a printable string at all times. And only use the binary layout at `Contact::mName`, in order to be able to split it into first and lastnames when required.

sergiohs84: Alright, the "fullname" already includes such space separator, which is good. Nevertheless, I don't see the advantage of the binary layout for the title. Please, review my next commit, @jgandres

PR #878, opened 2020-05-25
#15324. Improve performance in audio level monitor

alber2510:
Before merge notify apps developers to add binding if they need


src/rtcModule/webrtcPrivate.h
```diff
@@ -301,6 +304,8 @@ class Call: public ICall
     void updateAvFlags(karere::Id userid, uint32_t clientid, karere::AvFlags flags);
     bool isCaller(karere::Id userid, uint32_t clientid);
     void changeVideoInDevice();
+    virtual bool isAudioMonitorEnabled() const override;
+    virtual void enableAdioMonitor(bool enable) override;
```
sergiohs84, line 308:
```suggestion
    virtual void enableAudioLevelMonitor(bool enable) override;
```
`virtual` should not be here, but `override`

src/rtcModule/webrtcPrivate.h
```diff
@@ -36,6 +36,8 @@ class AudioLevelMonitor : public webrtc::AudioTrackSinkInterface
     ISessionHandler &mSessionHandler;
     const Session &mSession;
     bool mAudioDetected = false;
+
+    bool isDisabledAudioLevelMonitor() const;
```
sergiohs84, line 40:
```suggestion
    bool isAudioLevelMonitorEnabled() const;
```

src/rtcModule/webrtcPrivate.h
```diff
@@ -301,6 +304,8 @@ class Call: public ICall
     void updateAvFlags(karere::Id userid, uint32_t clientid, karere::AvFlags flags);
     bool isCaller(karere::Id userid, uint32_t clientid);
     void changeVideoInDevice();
+    virtual bool isAudioMonitorEnabled() const override;
```
sergiohs84, line 307:
```suggestion
    bool isAudioLevelMonitorEnabled() const override;
```
`virtual` should not be here, but `override`

src/rtcModule/webrtcPrivate.h
```diff
@@ -233,6 +235,7 @@ class Call: public ICall
     bool mIsRingingOut = false;
     bool mHadRingAck = false;
     bool mRecovered = false;
+    bool mAudioLevelMonitor = true;
```
sergiohs84, line 238:
```suggestion
    bool mAudioLevelMonitorEnabled = true;
```

src/megachatapi_impl.cpp
```diff
@@ -4087,6 +4087,70 @@ int MegaChatApiImpl::getMaxVideoCallParticipants()
...
+    if (!call)
+    {
+        API_LOG_ERROR("isAudioMonitorEnabled - There is not any call associated to MegaChatCallHandler");
+        return false;
+    }
+
+    return call->isAudioMonitorEnabled();
+}
+
+void MegaChatApiImpl::enableAudioMonitor(bool enable, MegaChatHandle chatid)
```
sergiohs84, line 4122: Maybe this method should return the effective value. Otherwise, callers will need to check, with an additional call to `isAudioMonitorEnabled()`, if the call was successful or failed.

src/rtcModule/webrtc.cpp
```diff
@@ -4330,6 +4340,13 @@ void AudioLevelMonitor::OnData(const void *audio_data, int bits_per_sample, int
     }
 }
 
+bool AudioLevelMonitor::isDisabledAudioLevelMonitor() const
+{
```
sergiohs84, line 4344: I think this method should rely exclusively on the new flag. It will be initialized upon the creation of the call and used directly (no need for this method) in order to decide whether the audio level should be monitored or not in `AudioLevelMonitor::OnData()`.

src/rtcModule/webrtc.cpp
```diff
@@ -4288,9 +4298,7 @@ AudioLevelMonitor::AudioLevelMonitor(const Session &session, ISessionHandler &se
 
 void AudioLevelMonitor::OnData(const void *audio_data, int bits_per_sample, int /*sample_rate*/, size_t number_of_channels, size_t number_of_frames)
 {
-    assert(bits_per_sample == 16);
-    time_t nowTime = time(NULL);
-    if (!mSession.receivedAv().audio())
+    if (isDisabledAudioLevelMonitor())
```
sergiohs84, line 4301:
```suggestion
    if (!mSession.call().isAudioMonitorEnabled() || !mSession.receivedAv().audio())
```

src/megachatapi.h
```diff
@@ -196,6 +196,9 @@ class MegaChatSession
     /**
      * @brief Returns if audio is detected for this session
      *
+     * @note This value will be always false for call in 1to1 chatrooms, when audio flag is disabled
+     * and in a group call when participants will be less than 7
+     *
```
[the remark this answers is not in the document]
alber2510: Fixed

src/rtcModule/webrtcPrivate.h
alber2510: Fixed

src/megachatapi_impl.cpp

src/rtcModule/webrtcPrivate.h
alber2510: Fixed

src/rtcModule/webrtc.cpp

src/megachatapi.h
```diff
@@ -4660,6 +4663,33 @@ class MegaChatApi
...
+     * @param chatid MegaChatHandle that identifies the chat room from we want know if audio level monitor is disabled
+     * @return true if audio level monitor is enabled
+     */
+    bool isAudioLevelMonitorEnabled(MegaChatHandle chatid);
+
+    /**
+     * @brief Enable or disable audio level monitor
+     *
+     * It's false by default and it's app responsability to enable it
```
sergiohs84, line 4682: Missing documentation about the request and its params :)

src/megachatapi_impl.cpp
```diff
@@ -1813,6 +1813,39 @@ void MegaChatApiImpl::sendPendingRequests()
...
             break;
         }
+        case MegaChatRequest::TYPE_ENABLE_AUDIO_LEVEL_MONITOR:
+        {
+            handle chatid = request->getChatHandle();
+            bool enable = request->getFlag();
+            if (chatid == MEGACHAT_INVALID_HANDLE)
+            {
+                API_LOG_ERROR("MegaChatRequest::TYPE_ENABLE_AUDIO_LEVEL_MONITOR - Invalid chatid");
+                errorCode = MegaChatError::ERROR_NOENT;
```
sergiohs84, line 1823: I'd say an invalid handle is an invalid arguments (`ERROR_ARGS`), but it's true that we already have `ERROR_NOENT` in multiple cases like this already. Up to you :)

src/megachatapi.h
alber2510: Fixed

src/megachatapi_impl.cpp
alber2510: I use `ERROR_NOENT` because I observed that it was used in more case but I think it's more correct `ERROR_ARGS`
I've set `ERROR_ARGS`

src/rtcModule/webrtcPrivate.h
alber2510: Fixed

src/rtcModule/webrtc.cpp
alber2510: Implement in other way. Solved

PR #880, opened 2020-06-01
#14744. Improve webrtc stats

alber2510:


src/chatClient.cpp
```diff
@@ -83,6 +83,7 @@ Client::Client(::mega::MegaApi& sdk, WebsocketsIO *websocketsIO, IApp& aApp, con
           chats(new ChatRoomList(*this)),
           mPresencedClient(&api, this, *this, caps)
 {
+    setlocale(LC_NUMERIC, "en_GB");
```
sergiohs84, line 86: Where this locale come from? What is its purpose?
The locale of the system might be different, and the user may have chosen a different one of his/her preference.

alber2510: This locale set the dot as decimal separator. In Spanish local, we use the comma and when we send a vector of floats in json, we use the comma to distinguish numbers
Only modify how numeric as show but before merging we have to check if this change affect to apps.
I've observer the problem in rtcstats.cpp method `decToString` but it can appear in other part of the code
If you think that it's more clean modify how `decToString` is working, I've tested a solution with that modification

sergiohs84: I think it's unsafe to set a locale without restoring it afterwards. The RapidJson library used to have an issue https://github.com/Tencent/rapidjson/issues/72. They solved it along with other improvements for floats/doubles: https://github.com/Tencent/rapidjson/pull/97, so the printing of floats is not anymore based on `snprintf()` (which honors locale).
A similar approach could be followed by `decToString`, I think it will be safer ;)

PR #887, opened 2020-06-16
#10235. Retention time for history improvements

jgandres:


src/chatd.cpp
```diff
@@ -4831,6 +4941,7 @@ void Chat::msgIncomingAfterDecrypt(bool isNew, bool isLocal, Message& msg, Idx i
         mLastHistDecryptCount++;
     }
 
+    bool checkRetentionHist = mOldestIdxInDb == CHATD_IDX_INVALID ;
```
sergiohs84, line 4944:
```suggestion
    bool checkRetentionHist = mOldestIdxInDb == CHATD_IDX_INVALID;
```

src/chatd.cpp
```diff
@@ -4421,6 +4490,47 @@ void Chat::handleRetentionTime()
...
+    if (lownum() == mOldestIdxInDb)
+    {
+        idx = CHATD_IDX_INVALID;
+        return;
+    }
+
+    CALL_DB(getIdxByRetentionTime, expireRetentionTs, idx);
+}
+
+time_t Chat::nextRetentionHistCheck(bool updateTimer)
```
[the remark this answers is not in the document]
jgandres: Sorry, this method was refactorized, and I forgot to reestablish the original condition.
We don't have to call updateRetentionCheckPeriod if updateTimer is false, to avoid an infitite loop
when handleRetentionTime upon timeour expiration.

Commit uploaded: f50f91c33b0e645ce04b0212e134e092cfc4ce3b

sergiohs84:
> karere-native-linux-PRs FAILED (1424) -1

It failed because it requires the homonymous branch of the SDK, and this PR does not have an homologous (but feature/retention-history does)

src/chatd.h
```diff
@@ -1439,6 +1458,12 @@ class Client : public karere::DeleteTrackable
     promise::Promise<void> sendKeepalive();
     void sendEcho();
 
+    /** Handler of the timeout for retention history checks */
+    megaHandle mRetentionTimer;
+
+    /** Period in seconds until next check of retention history for all chats, or zero (disabled) */
+    uint32_t mRetentionCheckPeriod;
```
alber2510, line 1465: With the last change it's not a period, it's the time stamp for next retention time. Probably other name improve the readability, per example mNextRetentionTime

src/chatd.cpp
```diff
@@ -362,6 +364,73 @@ void Client::setLastMsgTs(Id userid, ::mega::m_time_t lastMsgTs)
...
+    if (resetPeriod)
+    {
+        mRetentionCheckPeriod = 0;
+        CHATD_LOG_DEBUG("retention history check period reset");
+    }
+}
+
+void Client::setRetentionTimer()
+{
+    time_t nextCheck = mRetentionCheckPeriod - time(nullptr);
```
alber2510, line 400: Maybe retentionPeriod or similar improve readability

PR #890, opened 2020-06-23
#14875. Boringssl crashes

jgandres:


src/net/libwebsocketsIO.cpp
```diff
@@ -355,6 +361,11 @@ int LibwebsocketsClient::wsCallback(struct lws *wsi, enum lws_callback_reasons r
             {
                 X509_STORE_CTX_set_error((X509_STORE_CTX*)user, X509_V_OK);
             }
+            else
+            {
```
sergiohs84, line 365: In case `check_public_key()` returns false, we may want to send an event to stats server in order to record the incidence of the issue.

src/net/libwebsocketsIO.cpp
```diff
@@ -355,6 +361,11 @@ int LibwebsocketsClient::wsCallback(struct lws *wsi, enum lws_callback_reasons r
...
+                X509_STORE_CTX_set_error((X509_STORE_CTX*)user, X509_V_ERR_UNSPECIFIED);
```
sergiohs84, line 366: Maybe it's better to use a more specific error code here, since probably OpenSSL will use this `X509_V_ERR_UNSPECIFIED` for other reasons.
I think that `X509_V_ERR_APPLICATION_VERIFICATION` could be a good one.
> application verification failure 
>    an application specific error. This will never be returned unless explicitly set by an application.

src/net/libwebsocketsIO.cpp
```diff
@@ -321,6 +321,12 @@ static bool check_public_key(X509_STORE_CTX* ctx)
     EVP_PKEY* evp;
     if ((evp = X509_PUBKEY_get(X509_get_X509_PUBKEY(X509_STORE_CTX_get0_cert(ctx)))))
     {
+        if (EVP_PKEY_id(evp) != EVP_PKEY_RSA)
```
sergiohs84, line 324: This protection should be added to the homologous piece of code in the SDK, `cert_verify_callback()`. It is not currently in use for connections to chat-servers (it used to be, when DNS resolution was entirely done by the SDK and it checked the IP was reachable), but to API servers, and they might be affected by the same issue, so it deserves a patch too

src/net/libwebsocketsIO.cpp
```diff
@@ -355,6 +361,11 @@ int LibwebsocketsClient::wsCallback(struct lws *wsi, enum lws_callback_reasons r
             {
                 X509_STORE_CTX_set_error((X509_STORE_CTX*)user, X509_V_OK);
             }
+            else
+            {
+                X509_STORE_CTX_set_error((X509_STORE_CTX*)user, X509_V_ERR_UNSPECIFIED);
+                return 1;
```
sergiohs84, line 367: Usually this method returns `-1` upon error, and `0` when the reason has no specific handler. Since it's a critical zone, I'd double-check that, by returning `1`, libwebsocket will not reuse the failed cert in any way, but reset the SSL negotiation from the scratch.
A possible way to check is that, once this piece of code is executed, the connection is finally established to the right IP address (aka. chatd/presenced).

src/net/libwebsocketsIO.cpp
jgandres: Done in PR #2069

src/net/libwebsocketsIO.cpp
jgandres: I'm not sure If it's possible, because wsCallback is a static method and when we retrieve LibwebsocketsClient via user pointer, it seems that it's content it's not valid.

sergiohs84: I see. OK, let's keep it as it is by now :)

sergiohs84:
@jgandres please, can you prepare a hotfix PR for v2.3.5 too? We should patch it for upcoming app releases, since this one is a critical bugfix that should make into production ASAP. Thanks

PR #900, opened 2020-07-07
#15847. Improve qtApp GUI for calls

alber2510:
Add:
 - Number of video participants
 - Enable disable audio level monitor
 - Show participant that is speaking
 - Show if participant has enable or disable audio flag


sergiohs84:
Unfortunately, the latest changes use `QPalette::PlaceholderText`, which was introduced in Qt 5.12, and both Jenkins and some of the devs are still on 5.9.5 (Ubuntu's newest upgradable version). What do you suggest? Would it be possible to use a UI component that is supported by the old version, or should we update Jenkins and dev environments to a newer version?

alber2510:
I'll look for a other component compatible with older versions

PR #916, opened 2020-07-28
#16552. Check if mClient is valid at closeChatPreview

alber2510:


src/megachatapi_impl.cpp
```diff
@@ -3460,9 +3460,12 @@ void MegaChatApiImpl::closeChatRoom(MegaChatHandle chatid, MegaChatRoomListener
...
 void MegaChatApiImpl::closeChatPreview(MegaChatHandle chatid)
 {
-    sdkMutex.lock();
-    mClient->chats->removeRoomPreview(chatid);
-    sdkMutex.unlock();
+    SdkMutexGuard g(sdkMutex);
+
+    if (!mClient)
+        return;
+
```
sergiohs84, line 3467: I think the test doesn't need mutual exclusion here
```suggestion
    if (!mClient)
        return;
        
    SdkMutexGuard g(sdkMutex);

```

PR #917, opened 2020-07-29
#16483. Clean pending reaction from queue upon REJECT from chatd

jgandres:


src/chatd.cpp
```diff
@@ -2165,6 +2165,12 @@ void Connection::execCommand(const StaticBuffer& buf)
                 CHATDS_LOG_WARNING("%s: recv REJECT of %s: id='%s', reason: %hu",
                     ID_CSTR(chatid), Command::opcodeToStr(op), ID_CSTR(id), reason);
                 auto& chat = mChatdClient.chats(chatid);
+                if (op == OP_ADDREACTION || op == OP_DELREACTION)
+                {
+                    chat.onReactionReject(id);
+                    break;
+                }
```
sergiohs84, line 2172: The following `if` could become an `else if` now, couldn't it?

jgandres: Adjusted in commit: fcdaee475abafb7540c9d8964946bfbb8ae7bf14

PR #925, opened 2020-08-18
CHT-415. Update for building megaclc on windows without webrtc

mattw-mega:
Also added the startUpload command

### DIFF
--- a/src/net/libwebsocketsIO.cpp
+++ b/src/net/libwebsocketsIO.cpp
@@ -282,7 +282,7 @@ void LibwebsocketsClient::resetOutputBuffer()
 #define EVP_PKEY_get0_RSA(_pkey_) ((_pkey_)->pkey.rsa)
 #endif
 
-#ifdef USE_WEBRTC
+#ifndef WIN32
 const BIGNUM *RSA_get0_n(const RSA *rsa)
 {
 #if (OPENSSL_VERSION_NUMBER < 0x10100000L)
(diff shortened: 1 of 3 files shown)


src/net/libwebsocketsIO.cpp
```diff
@@ -282,6 +282,7 @@ void LibwebsocketsClient::resetOutputBuffer()
 #define EVP_PKEY_get0_RSA(_pkey_) ((_pkey_)->pkey.rsa)
 #endif
 
+#ifdef USE_WEBRTC
```
sergiohs84, line 285: I think these methods actually belongs to OpenSSL. When building with WebRTC, the crypto library is replaced by BoringSSL, which is already part of WebRTC library, and it's used for the SDK as well, as replacement of OpenSSL. However, without WebRTC, I think those methods should be kept.
What is exactly the issues with them on Windows?

sergiohs84: Actually, this change breaks building on Linux. 
image
The method below, ` static bool check_public_key(X509_STORE_CTX* ctx)` needs them.

mattw-mega: On windows the symbol is doubly defined - one of the 3rdParty libraries already has them.   I'll change it to #ifndef WIN32 then, thanks

sergiohs84 (approved):

PR #948, opened 2020-09-20
CHT-425: Feature/giphy support

ms-mega:


src/chatdMsg.h
```diff
@@ -513,7 +513,8 @@ class Message: public Buffer
     {
         kInvalid              = 0xff,
         kRichLink             = 0x00,
-        kGeoLocation          = 0x01
+        kGeoLocation          = 0x01,
```
sergiohs84, line 516: Do you know the purpose of the `0x02`?? it would be good to add it here, just for documentation. Maybe Lyubo can tell you.

src/megachatapi.h
```diff
@@ -3961,6 +4052,39 @@ class MegaChatApi
...
+     * lower too (and the next call to MegaChatApi::loadMessages will fetch messages from server).
+     *
+     * When there are no more history available from the reported source of messages
+     * (local / remote), or when the requested \c count has been already loaded,
+     * the callback  MegaChatNodeHistoryListener::onAttachmentLoaded will be called with a NULL message.
+     *
+     * @param chatid MegaChatHandle that identifies the chat room
+     * @param count The number of requested messages to load.
+     *
+     * @return Return the source of the messages that is going to be fetched. The possible values are:
```
sergiohs84, line 4078: It doesn't look consistent. Please, review the documentation

src/megachatapi_impl.cpp
```diff
@@ -9815,6 +9925,12 @@ const MegaChatContainsMeta* JSonUtils::parseContainsMeta(const char *json, uint8
                 containsMeta->setGeolocation(geolocation);
                 break;
             }
+            case MegaChatContainsMeta::CONTAINS_META_GIPHY:
+            {
+                MegaChatGiphy* giphy = parseGiphy(document);
```
sergiohs84, line 9930: Despite the old code doesn't use smart pointers here, maybe it's time to return a `unique_ptr<>` here? so the ownership of the object is easier to track.

src/megachatapi_impl.h
```diff
@@ -1322,25 +1349,29 @@ class MegaChatContainsMetaPrivate : public MegaChatContainsMeta
...
     void setGeolocation(MegaChatGeolocation *geolocation);
     void setTextMessage(const std::string &text);
+    void setGiphy(MegaChatGiphy *giphy);
 
 protected:
     int mType = MegaChatContainsMeta::CONTAINS_META_INVALID;
     std::string mText;
     MegaChatRichPreview *mRichPreview = NULL;
     MegaChatGeolocation *mGeolocation = NULL;
+    MegaChatGiphy *mGiphy = nullptr;
```
sergiohs84, line 1365: You may want to use a smart pointer here, instead of a raw pointer.

sergiohs84 (commented):
Just some comments from a quick glance ;)

src/chatdMsg.h
ms-mega: guess 0x02 is reserved for something or might just be an oversight. Will check.

ms-mega:
trigger compilation

ms-mega:
trigger compilation

src/megachatapi_impl.cpp
```diff
@@ -9933,6 +10044,98 @@ MegaChatGeolocation *JSonUtils::parseGeolocation(rapidjson::Document &document)
...
+    {
+        auto giphyheightString = giphyheightIterator->value.GetString();
+        giphyHeight = atoi(giphyheightString);
+    }
+    else
+    {
+        API_LOG_ERROR("parseGiphy: invalid JSON struct - \"h\" field not found");
+        return std::unique_ptr<MegaChatGiphy>(nullptr);
+    }
+    return std::unique_ptr<MegaChatGiphyPrivate> (new MegaChatGiphyPrivate(mp4srcString, webpsrcString, mp4Size, webpSize, giphyWidth, giphyHeight, giphyTitle));
```
ms-mega, line 10136: It's C++11. So, can't use make_unique yet.

aabellagm:
trigger compilation

src/megachatapi_impl.cpp
sergiohs84: You can use `::mega::make_unique<>` in the meantime :smiley:

src/megachatapi_impl.cpp
```diff
@@ -9148,7 +9148,7 @@ const MegaChatGeolocation *MegaChatContainsMetaPrivate::getGeolocation() const
 
 const MegaChatGiphy* MegaChatContainsMetaPrivate::getGiphy() const
 {
-    return mGiphy;
+    return mGiphy ? mGiphy.get() : nullptr;
```
sergiohs84, line 9151: Cannot simply return `mGiphy.get()` in all cases
> Returns a pointer to the managed object or nullptr if no object is owned. 
```suggestion
    return mGiphy.get();
```
https://en.cppreference.com/w/cpp/memory/unique_ptr/get

src/megachatapi_impl.cpp
```diff
@@ -9194,22 +9194,17 @@ void MegaChatContainsMetaPrivate::setTextMessage(const string &text)
     mText = text;
 }
 
-void MegaChatContainsMetaPrivate::setGiphy(MegaChatGiphy* giphy)
+void MegaChatContainsMetaPrivate::setGiphy(std::unique_ptr<MegaChatGiphy>& giphy)
```
sergiohs84, line 9197: The idea is to receive the ownership directly, instead of a reference + move. That way, it's clear that the caller gives the ownership (you would need to `setGiphy(move(giphy))` and here you simply do `mGiphy.reset(giphy)` :)

sergiohs84 (changes requested):
See my comments, please.

src/megachatapi_impl.cpp
ms-mega: made the suggested changes pertaining to ownership. 
I think we can't do a mGiphy.reset(giphy). std::move() seems to be the right operation.

sergiohs84 (dismissed):

sergiohs84 (approved):

PR #957, opened 2020-09-29
CHT-428. Uncomment reactions limitations code

jgandres:
Merge this PR at the same time Webclient add this enhancement

### DIFF
--- a/src/megachatapi_impl.cpp
+++ b/src/megachatapi_impl.cpp
@@ -2022,15 +2022,14 @@ void MegaChatApiImpl::sendPendingRequests()
             bool hasReacted = msg.hasReacted(reaction, mClient->myHandle());
             if (request->getFlag())
             {
-//                Uncomment this block upon release of reaction limitations
-//                // check if max number of reactions has been reached
-//                int res = msg.allowReact(mClient->myHandle(), reaction);
-//                if (res != 0)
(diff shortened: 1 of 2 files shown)


src/megachatapi_impl.cpp
```diff
@@ -2022,15 +2022,14 @@ void MegaChatApiImpl::sendPendingRequests()
...
-//                    request->setNumber(res);
-//                    errorCode = MegaChatError::ERROR_NOENT;
-//                    break;
-//                }
+                // check if max number of reactions has been reached
+                int res = msg.allowReact(mClient->myHandle(), reaction);
+                if (res != 0)
+                {
+                    request->setNumber(res);
+                    errorCode = MegaChatError::ERROR_NOENT;
```
sergiohs84: Could we return a new error code that doesn't clash with this one, already used for non existing `msgid`??
Perhaps a `ERROR_TOOMANY` (we can define it, in case it doesn't exist yet).

jgandres: I returned ERROR_NOENT as ticket specified, but there's no problem to change it.

jgandres: Fixed in db87993eda1bbe0cc1cc71fcc19de463e7e9e252

